PR #85, opened 2022-08-11
Adding semester status request form 

sszajbely:
Solves #77 
Form for collegists to update their status at the beginning of every semester. Users with deactivated status in the previous semester will not be able to alter their status, and their status in the new semester will automatically be deactivated.
Users filling in the form will be able to choose between the semester statuses active, passive and deactivated and between the residence statuses resident and extern. This will alter these for the current semester.
There is an email sent out automatically on the first of every september and february asking the collegists to fill out the form, if they fail to do so until the 15th of the month, their status will be set to deactivated automatically.
Possible enhancements: 
- Show the form in the given period in the menu bar (now it is only accessible through an url or the email.  
- Send out emails to the workshop secretaries in which we ask them to send out an email to all workshop members.


resources/views/emails/status_statement_request.blade.php
```diff
@@ -0,0 +1,10 @@
+@component('mail::message')
+<h1>@lang('mail.dear') {{ $recipent??'asd' }}!</h1>
```
kdmnk, line 2: delete asd

PR #105, opened 2022-09-20
Create new roles based on the new regulations of Eotvos Collegium

sszajbely:
New roles:
- Economic vice president (formerly economic committee leader)
- Scientific vice president  (formerly scientific committee leader)
- Students' council secretary
- Trustees bord member
- Referent for every committee

There is no economic and scientific committee anymore, so I removed the members and leaders of them.

Additional changes:
- Modified permission policies
- Fix in Semester model: the parts are saved as strings, but the current() method tried to search them with ints.


app/Models/Semester.php
```diff
@@ -201,10 +201,10 @@ public static function current(): Semester
         if (! Cache::get('semester.current.'.$today)) {
             $now = Carbon::now();
             if ($now->month >= self::START_OF_SPRING_SEMESTER && $now->month <= self::END_OF_SPRING_SEMESTER) {
-                $part = 2;
+                $part = "2"; // Because of some strange reaseon, the part is stored as a string.
                 $year = $now->year - 1;
             } else {
-                $part = 1;
+                $part = "1";
```
dokatrifork: I'd prefer changing the representation instead

dokatrifork: And please add this in a different PR.

app/Models/Semester.php
```diff
@@ -286,7 +286,7 @@ public static function getOrCreate($year, $part): Semester
         }
         $semester = Semester::firstOrCreate([
             'year' => $year,
-            'part' => $part,
+            'part' => (string) $part,
```
dokatrifork: See above

app/Policies/CheckoutPolicy.php
```diff
@@ -38,14 +38,14 @@ public function addPayment(User $user, Checkout $checkout): bool
 
     public function addKKTNetreg(User $user): bool
     {
-        return $user->hasRole(Role::STUDENT_COUNCIL, Role::ECONOMIC_MEMBER)
-            || $user->hasRole(Role::STUDENT_COUNCIL, Role::ECONOMIC_LEADER);
+        return $user->hasRole(Role::STUDENT_COUNCIL, Role::ECONOMIC_VICE_PRESIDENT)
+            || $user->hasRole(Role::STUDENT_COUNCIL, Role::ECONOMIC_VICE_PRESIDENT);
```
dokatrifork: The same thing twice

app/Policies/TransactionPolicy.php
```diff
@@ -25,10 +25,9 @@ public function delete(User $user, Transaction $transaction): bool
         }
 
         if ($transaction->checkout->name == Checkout::studentsCouncil()->name) {
-            return $user->hasRole(Role::STUDENT_COUNCIL, Role::ECONOMIC_LEADER)
-                || (
-                    $user->hasRole(Role::STUDENT_COUNCIL, Role::ECONOMIC_MEMBER)
-                    && $transaction->receiver_id == $user->id
+            return $user->hasRole(Role::STUDENT_COUNCIL, Role::ECONOMIC_VICE_PRESIDENT)
+                || (//is this okay this way?
```
dokatrifork: We needed the check like this to ensure that economic members could only delete the transaction they received. You can just leave this as a economic_vice_president check. (Are we sure that the president will not need any help collecting the transactions?)

app/Policies/UserPolicy.php
```diff
@@ -28,9 +28,9 @@ public function viewAny(User $user): bool
     {
         return
             $user->hasAnyRoleBase([
-                Role::STAFF, Role::SECRETARY, Role::DIRECTOR, Role::WORKSHOP_ADMINISTRATOR, Role::WORKSHOP_LEADER
+                Role::STAFF, Role::SECRETARY, Role::DIRECTOR, Role::WORKSHOP_ADMINISTRATOR, Role::WORKSHOP_LEADER, Role::STUDENT_COUNCIL_SECRETARY
             ])
-            || $user->isPresident();
+            || $user->isPresident() || $user->hasRole(Role::STUDENT_COUNCIL, Role::SCIENCE_VICE_PRESIDENT);
```
dokatrifork: You might want to add the other vice president also. Create an isStudentCouncilLeaders (or similar) function for this.

app/Policies/UserPolicy.php
```diff
@@ -122,19 +125,36 @@ public function updateAnyPermission(User $user, User $target, Role $role = null)
...
+        if ($role->name == Role::WORKSHOP_ADMINISTRATOR) {
+            return $user->hasRoleBase(Role::WORKSHOP_LEADER) || $user->hasRole(Role::STUDENT_COUNCIL_SECRETARY) || $user->hasRole(Role::SECRETARY) || $user->hasRole(Role::STUDENT_COUNCIL, Role::SCIENCE_VICE_PRESIDENT);
+        }
+
+        if ($role->name == Role::STUDENT_COUNCIL_SECRETARY) {
+            return $user->isStudentsCouncilSecretary();
+        }
+
+        if ($role->name == Role::BOARD_OF_TRUSTEES_MEMBER) {
+            return $user->isStudentsCouncilSecretary();
```
dokatrifork, line 149: Don't we want to add the president here?

app/Policies/UserPolicy.php
```diff
@@ -122,19 +125,36 @@ public function updateAnyPermission(User $user, User $target, Role $role = null)
...
+        if ($role->name == Role::WORKSHOP_LEADER) {
+            return $user->hasRole(Role::SECRETARY) || $user->hasRole(Role::DIRECTOR);
+        }
+
+        if ($role->name == Role::WORKSHOP_ADMINISTRATOR) {
+            return $user->hasRoleBase(Role::WORKSHOP_LEADER) || $user->hasRole(Role::STUDENT_COUNCIL_SECRETARY) || $user->hasRole(Role::SECRETARY) || $user->hasRole(Role::STUDENT_COUNCIL, Role::SCIENCE_VICE_PRESIDENT);
+        }
+
+        if ($role->name == Role::STUDENT_COUNCIL_SECRETARY) {
+            return $user->isStudentsCouncilSecretary();
```
dokatrifork, line 145: Don't we want to add the president here?

app/Policies/UserPolicy.php
```diff
@@ -47,7 +47,10 @@ public function view(User $user, User $target): bool
             if ($user->hasAnyRoleBase([Role::SECRETARY, Role::DIRECTOR])) {
                 return true;
             }
-            if ($user->isPresident()) {
+            if ($user->isPresident() || $user->hasRole(Role::STUDENT_COUNCIL, Role::SCIENCE_VICE_PRESIDENT)) {
```
dokatrifork: same as above

app/Policies/UserPolicy.php
```diff
@@ -156,14 +176,33 @@ public function updatePermission(User $user, User $target, Role $role, Workshop|
             return $user->roleWorkshops()->contains($object->id);
         }
 
+        if ($role->name == Role::WORKSHOP_LEADER) {
+            return $user->hasRole(Role::SECRETARY) || $user->hasRole(Role::DIRECTOR);
+        }
+
+        if ($role->name == Role::WORKSHOP_ADMINISTRATOR) {
+            return $user->hasRoleBase(Role::WORKSHOP_LEADER) || $user->hasRole(Role::STUDENT_COUNCIL_SECRETARY) || $user->hasRole(Role::SECRETARY) || $user->hasRole(Role::STUDENT_COUNCIL, Role::SCIENCE_VICE_PRESIDENT);
```
dokatrifork: Only the corresponding workshop leader should add the workshop administrator role. I have similar functionalities somewhere with a $user->roleWorkshops call or sth similar

app/Policies/UserPolicy.php
```diff
@@ -156,14 +176,33 @@ public function updatePermission(User $user, User $target, Role $role, Workshop|
...
+        }
+
+        if ($role->name == Role::STUDENT_COUNCIL_SECRETARY) {
+            return $user->isStudentsCouncilSecretary();
+        }
+
+        if ($role->name == Role::BOARD_OF_TRUSTEES_MEMBER) {
+            return $user->isStudentsCouncilSecretary();
+        }
```
dokatrifork, line 193: Same as above

database/migrations/2022_09_20_194258_change_roles_to_new_regulations.php
```diff
@@ -0,0 +1,47 @@
...
+    public function up()
+    {
+        DB::table('roles')->updateOrInsert(
+            ['name'=>'student-council-secretary'],
+            ['has_workshops' => 0, 'has_objects' => 0]
+        );
+        DB::table('roles')->updateOrInsert(
+            ['name'=>'student-council-secretary'],
+            [ 'has_workshops' => 0, 'has_objects' => 0]
+        );
```
dokatrifork: Same twice

database/migrations/2022_09_20_194258_change_roles_to_new_regulations.php
```diff
@@ -0,0 +1,47 @@
...
+        DB::table('role_objects')->where('name', 'science-leader')->delete();
+        DB::table('role_objects')->where('name', 'economic-member')->delete();
+        DB::table('role_objects')->where('name', 'science-member')->delete();
+        DB::table('role_objects')->updateOrInsert(['role_id'=> Role::firstWhere('name', 'student-council')->id, 'name' => 'science-vice-president'], []);
+        DB::table('role_objects')->updateOrInsert(['role_id'=> Role::firstWhere('name', 'student-council')->id, 'name' => 'economic-vice-president'], []);
+        DB::table('role_objects')->updateOrInsert(['role_id'=> Role::firstWhere('name', 'student-council')->id, 'name' => 'cultural-referent'], []);
+        DB::table('role_objects')->updateOrInsert(['role_id'=> Role::firstWhere('name', 'student-council')->id, 'name' => 'community-referent'], []);
+        DB::table('role_objects')->updateOrInsert(['role_id'=> Role::firstWhere('name', 'student-council')->id, 'name' => 'communication-referent'], []);
+        DB::table('role_objects')->updateOrInsert(['role_id'=> Role::firstWhere('name', 'student-council')->id, 'name' => 'sport-referent'], []);
```
dokatrifork, line 36: Are you sure these work? I think these get matched by the id and the new name. Even if so, you might want to use InsertOrIgnore to be more clear.

resources/lang/en/role.php
```diff
@@ -5,21 +5,26 @@
...
+    'community-referent' => 'Community committee referent',
     'cultural-committee' => 'Cultural Committee',
     'cultural-leader' => 'Cultural committee leader',
     'cultural-member' => 'Cultural committee member',
+    'cultural-referent' => 'Cultural committee referent',
     'de' => 'German',
     'director' => 'Director',
     'economic-committee' => 'Economic Committee',
     'economic-leader' => 'Economic committee leader',
     'economic-member' => 'Economic committee member',
```
dokatrifork: Delete the old ones (elsewhere also)

dokatrifork (changes requested):

app/Models/Semester.php
sszajbely: I'll create a new PR, I don't really understand why this is not working currently.

app/Policies/TransactionPolicy.php
sszajbely: In recent years anyone can pay KKT by bank transfer, so help shouldn't be needed.

app/Policies/UserPolicy.php
sszajbely: I don't think that they should be able to assign the board of trustees members.

app/Policies/UserPolicy.php
sszajbely: See above

app/Policies/UserPolicy.php
sszajbely: See above

database/migrations/2022_09_20_194258_change_roles_to_new_regulations.php
sszajbely: After a session of googling this was the method that I could come up with. insertOrIgnore just ignores that there exists another record like this and adds it regardless.
updateOrInsert works by looking at the first argument if there are matches corresponding to that input, and if there are updates the second argument. If there aren't any, the first and second input will be inserted combined.

resources/lang/en/role.php
sszajbely: I hoped you'd miss that :D

database/migrations/2022_09_20_194258_change_roles_to_new_regulations.php
```diff
@@ -0,0 +1,47 @@
...
```
kdmnk, line 36: @machiato32 next time use rename, this way the constraints fail when you delete a role.

PR #111, opened 2022-09-24
Optimization for role handling and querying 

kdmnk:
Modified hasRole() signature.
Optimized role queries with cache


kdmnk:
user list page load (for admin users):
before:
313 statements were executed, 278 of which were duplicated, 35 unique - 210ms
after:
20 statements were executed -60ms
:D

PR #112, opened 2022-09-24
Changed all occurances of role.economic-committee to role.economic-vice-president

### LINKED ISSUE #110: Rename role.economic-committe locale usages
For example, in the navbar, we still use locales regarding the old committee. Rename those

sszajbely:
#110


kdmnk:
Would not be better to display `student council checkout` in the navbar and titles?

kdmnk (approved):
see above

kdmnk (approved):

PR #118, opened 2022-09-25
Fixed alphabetical sorting in exported excel sheet

### LINKED ISSUE #109: Excel export ordering not right
It gives á, é and so on after all others.

sszajbely:
#109


kdmnk:
Is this working properly with the special Hungarian letters? I think we already did something with this issue.

kdmnk (approved):

PR #119, opened 2022-09-25
Lost commits in release branch

kdmnk:
These were in the release branch, which we are not using anymore.
Review what we need or already done from here - probably better to redo these in a new branch.


resources/lang/hu/registration.php
```diff
@@ -5,6 +5,8 @@
     'confirmpwd' => 'Jelszó megerősítése',
     'email' => 'E-mail cím',
     'forgotpwd' => 'Elfelejtetted a jelszót?',
+    'information' => 'A jelentkezéshez szükséges további információkat a regisztráció után lehet feltölteni.',
+    'information_tenant' => 'Ez az oldal a leendő collegistáknak szól. Ha vendégként szeretnél regisztrálni, kattints ide.',
```
sszajbely, line 9: Not in current version

resources/views/auth/register.blade.php
```diff
@@ -14,6 +14,9 @@
                     @foreach ($errors->all() as $error)
                         <blockquote class="error">{{ $error }}</blockquote>
                     @endforeach
+		@if($user_type == \App\Models\Role::COLLEGIST)
+		<blockquote><a href="{{route('register.guest')}}">@lang('registration.information_tenant')</a></blockquote>
+		@endif
```
sszajbely, line 19: Not in current version

PR #126, opened 2022-09-26
Status form fix

### LINKED ISSUE #106: Reminder for users on status form that they already set their status

sszajbely:


sszajbely:
@kdmnk can you try if it redirects users without the collegist role?

app/Http/Controllers/Secretariat/SemesterController.php
```diff
@@ -14,13 +14,13 @@
...
-    public static function isStatementAvailable()
+    public static function isStatementAvailable($user)
     {
-        $statement_event = EventTrigger::find(EventTrigger::INTERNET_ACTIVATION_SIGNAL)->date;
+        $statement_event = EventTrigger::find(EventTrigger::SEND_STATUS_STATEMENT_REQUEST)->date;
         $deadline_event = EventTrigger::find(EventTrigger::DEACTIVATE_STATUS_SIGNAL)->date;
         // If the deadline is closer than sending out the request, that means
         // the request has been already sent out.
-        return $deadline_event < $statement_event;
+        return $deadline_event < $statement_event || $user->getStatusIn(Semester::current())==SemesterStatus::INACTIVE;
```
kdmnk, line 23: Comment on the other PR to be aware of this.

routes/web.php
```diff
@@ -152,7 +152,8 @@
     Route::put('/rooms/update', [RoomController::class, 'updateResidents'])->name('rooms.update');
 
     /** Status update form */
-    Route::get('/secretariat/status-update', [SemesterController::class, 'showStatusUpdate'])->name('secretariat.status-update.show');
+    Route::get('/secretariat/status-update', [SemesterController::class, 'showStatusUpdate'])->name('secretariat.status-update.show')->withoutMiddleware([EnsureStatus::class]);
+    ;
```
kdmnk: ;

app/Http/Middleware/EnsureStatus.php
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure(\Illuminate\Http\Request): (\Illuminate\Http\Response|\Illuminate\Http\RedirectResponse)  $next
+     * @return \Illuminate\Http\Response|\Illuminate\Http\RedirectResponse
+     */
+    public function handle(Request $request, Closure $next)
+    {
+        if (!$request->is('logout') && !$request->routeIs('secretariat.status-update.*') && $request->user() && $request->user()->getStatusIn(Semester::current())==SemesterStatus::INACTIVE) {
```
kdmnk: add isCollegist()

app/Http/Controllers/Secretariat/SemesterController.php
```diff
@@ -32,10 +32,10 @@ public function showStatusUpdate()
         if ($user->getStatusIn(Semester::previous()->id) == SemesterStatus::DEACTIVATED) {
             abort(403);
         }
-        if (!self::isStatementAvailable()) {
+        if (!self::isStatementAvailable($user)) {
             abort(403);
         }
-        return view('secretariat.statuses.status_update_form');
+        return view('secretariat.statuses.status_update_form', ['user' => Auth::user()]);
```
kdmnk: we usually get this from the blades directly

app/Http/Middleware/EnsureStatus.php
kdmnk: And maybe this could be simpler if you just add all these to the isStatementAvailable function and call it here

app/Http/Controllers/Secretariat/SemesterController.php
```diff
@@ -32,10 +32,10 @@ public function showStatusUpdate()
         if ($user->getStatusIn(Semester::previous()->id) == SemesterStatus::DEACTIVATED) {
             abort(403);
         }
-        if (!self::isStatementAvailable()) {
+        if (!self::isStatementAvailable($user)) {
             abort(403);
```
kdmnk: return redirect home with a session message that the status has been set already

app/Http/Middleware/EnsureStatus.php
kdmnk: You could add this to the routes as a withoutmiddleware for the logout.
Also, rename this file for a more general-purpose one where we can have more redirection rules in the future.

kdmnk (changes requested):
See comments

sszajbely:
#106

app/Http/Middleware/EnsureStatus.php
sszajbely: The `withoutMiddleware` function didn't really work, I just forgot to delete it in the code. And I don't think that we should be adding this to the `isStatementAvailable `function as these are not checking if it's available.

resources/lang/en/secretariat.php
```diff
@@ -6,6 +6,8 @@
     'semester_status' => 'Semester status',
     'semester_status_info' => 'Active - You will be an active member of the Collegium<br>Passive - You will be on a foregin semester or are passive in the university<br>Deactivated - You will not be an active member of the Collegium',
     'status' => 'Status',
+    'status_set' => 'Your status for this semester has already been set, it is',
+    'status_statement_not_available' => 'The status update form is not available. You have already set your status and the deadline to set it has passed.',
```
kdmnk, line 10: I think it's an 'or', not an 'and'

resources/views/secretariat/statuses/status_update_form.blade.php
```diff
@@ -4,6 +4,9 @@
 @endsection
 
 @section('content')
+@php
+    $user = \Illuminate\Support\Facades\Auth::user();
```
kdmnk: We can just use Auth::user() below.

resources/lang/en/secretariat.php
sszajbely: The status form is open if the deadline hasn't passed OR if the status is not set. So the negation becomes AND

resources/views/secretariat/statuses/status_update_form.blade.php
sszajbely: I use it two times below

kdmnk: But you do not need this as a variable

resources/lang/en/secretariat.php
kdmnk: the status form is open if the deadline has not passed AND the status is not set :D

kdmnk: Oh, you did not make the check like this.

kdmnk: Would not it make more sense?

resources/views/secretariat/statuses/status_update_form.blade.php
sszajbely: Isn't it faster to store it as a variable and use it afterwards?

resources/lang/en/secretariat.php
sszajbely: It does not make too much difference as everyone's status will be set to DEACTIVATED if they didn't set it themselves. I think it is okay to allow people to set their status multiple times before the deadline, for example if they messed it up the first time.

resources/views/secretariat/statuses/status_update_form.blade.php
kdmnk: I'm not familiar with the technical details, but I assume that the current user is stored in the session or somewhere else. Almost every code piece uses it. Or even if it is not, we do not have it in a variable in other blades, so for now, stick with that.

resources/lang/en/secretariat.php
kdmnk: Okay

kdmnk (approved):

PR #131, opened 2022-09-28
Improvements and fixes for the room assignment module

### LINKED ISSUE #70: Pause router pinging on debug mode
We get a lot of unnecessary emails when the routers seem unavailable while the page is in debug mode.

### LINKED ISSUE #123: Add option to select tenants in the room assignment form
If they are still tenants (check the `tenant_until` field)

sszajbely:
- Active tenants are displayed in the selectors
- Room assignment blade uses functions to get users and color of the room
- President can assign tenant role: I added this because I don't want to add externs to the query, but right now they cannot be added to the room assignment. So the new policy should be to add the tenant right to the externs living in the building.
- Made separate page for modifying the room assignment. This should make the loading faster.
- Added email that is sent out to the residents of the room if the router is not available.


app/Http/Controllers/Dormitory/RoomController.php
```diff
@@ -18,10 +21,26 @@ public function index()
     {
         $this->authorize('viewAny', Room::class);
         $users=User::active()->resident()->get();
-
+        $tenants=User::role(Role::getRole(Role::TENANT), null)->whereHas('personalInformation', function ($q) {
```
kdmnk: User::role(Role::TENANT)->... is enough

app/Http/Controllers/Dormitory/RoomController.php
```diff
@@ -18,10 +21,26 @@ public function index()
...
+            $q->where('tenant_until', '>', now());
+        })->get();
```
kdmnk: Add an currentTenants abstract function to the User model

app/Models/Room.php
```diff
@@ -16,10 +16,6 @@ class Room extends Model
 
     protected $fillable = ['name', 'capacity'];
 
-    protected $with = ['users'];
-
-    protected $withCount = ['users'];
-
```
kdmnk: I added these to optimize the queries... Why did you delete it?

app/Models/Router.php
```diff
@@ -53,6 +53,13 @@ public function sendWarning()
             foreach ($internet_admins as $admin) {
                 Mail::to($admin)->queue(new \App\Mail\RouterWarning($admin, $this));
             }
+            $room=Room::firstWhere('name', $this->room);
+            if ($room!=null) {
+                $residents=$room->users()->get();
+                foreach ($residents as $resident) {
```
kdmnk: foreach($room->residents as ...

app/Models/Router.php
```diff
@@ -53,6 +53,13 @@ public function sendWarning()
             foreach ($internet_admins as $admin) {
```
kdmnk, line 53: Also please do https://github.com/EotvosCollegium/mars/issues/70 also, while you are here

app/Policies/UserPolicy.php
```diff
@@ -137,6 +137,10 @@ public function updateAnyPermission(User $user, User $target, Role $role = null)
             ]);
         }
 
+        if ($role->name == Role::TENANT) {
+            return $user->hasRole([Role::STAFF, Role::STUDENT_COUNCIL => Role::PRESIDENT]);
+        }
```
kdmnk, line 142: Then, you should add the president to the view policy as well (if not there already)

resources/views/dormitory/rooms/app.blade.php
```diff
@@ -124,6 +124,14 @@
                         array_push($colorArray, [ "id" => $room->name, "color" => $color]);
                     }
                     $roomColors=collect($colorArray);
+
+                    function getRoomColor(string|int $roomNumber, $roomColors) {
+                        return $roomColors->where('id',$roomNumber)->pluck('color')->join(',');
+                    }
+
+                    function getRoomResidents(string|int $roomNumber, \Illuminate\Database\Eloquent\Collection $rooms){
+                        return $rooms->firstWhere('name', $roomNumber)->users()->get()->pluck('name')->join(';');
```
kdmnk: Do not add functions to blade files, please :D

kdmnk (changes requested):

app/Models/Room.php
sszajbely: It doesn't work. If I type $room->users it returns the first user, not all of them. Just $room->users()->get() works for some reason

resources/views/dormitory/rooms/app.blade.php
sszajbely: And where should they be then? Like this it does not look like shit :D

app/Models/Room.php
kdmnk: - still, you've multiplied tenfold the number of queries :P
- I checked, the relation was bad from those rooms' side. Please modify it as `return $this->hasMany(User::class, 'room', 'name');`

kdmnk:
I also noticed some other things:
- RoomController - updateResidents:
instead of:
```
        foreach ($users as $user) {
            $user->update(['room' => null]);
        }
```
`$users->update(['room' => null]);` is enough

- Currently, when a user has been selected for two different users, the system will only store it for the latter room without any warnings.
- Actually, I would prefer to have this as a livewire page, this way we can also display only the users without any room for selection. Do you want to do that?

resources/views/dormitory/rooms/app.blade.php
kdmnk: I'd make an attribute getter for the room model that returns the colour accordingly and the names as a string. Then,  you can make one big array with different attributes in the controller and pass it to the view. I want to avoid using PHP in blades as much as possible

resources/views/dormitory/rooms/app.blade.php
```diff
@@ -145,46 +153,46 @@
...
+                        <path id="room224" style="fill: {{getRoomColor(224, $roomColors)}}" data-info="{{ getRoomResidents(224, $rooms) }}" class="st1" d="M174.9,525.5h51.4l-0.2,92l-51.4,0.4L174.9,525.5z"/>
+                        <path id="room223" style="fill: {{getRoomColor(223, $roomColors)}}" data-info="{{ getRoomResidents(223, $rooms) }}" class="st1" d="M228.6,525.5l43.4,0.4l0.2,91.9h-44.2v-92.2H228.6z"/>
+                        <path id="room222" style="fill: {{getRoomColor(222, $roomColors)}}" data-info="{{ getRoomResidents(222, $rooms) }}" class="st1" d="M274.5,525.5l47.9,0.2l0.6,91.9l-48.5,0.2L274.5,525.5z"/>
+                        <path id="room221" style="fill: {{getRoomColor(221, $roomColors)}}" data-info="{{ getRoomResidents(221, $rooms) }}" class="st1" d="M324.4,525.5h46.4v92l-46.4,0.2V525.5z"/>
+                        <path id="room220" style="fill: {{getRoomColor(220, $roomColors)}}" data-info="{{ getRoomResidents(220, $rooms) }}" class="st1" d="M372.5,525.5l51.1,0.2l-0.4,92h-50.5L372.5,525.5z"/>
```
kdmnk: Instead of the style, use the materialize colour classes (https://materializecss.github.io/materialize/color.html), also you can get it from the model.
For the coordinates (and every other variable attribute), paste them into a new file (like we did with the countries), and you can also load them in the controller.
This way, you can make this in a foreach.

sszajbely:
> I also noticed some other things:
> 
> * RoomController - updateResidents:
>   instead of:
> 
> ```
>         foreach ($users as $user) {
>             $user->update(['room' => null]);
>         }
> ```
> 
> `$users->update(['room' => null]);` is enough

It doesn't seem to work for me. I get an error: `Method Illuminate\Database\Eloquent\Collection::update does not exist.`
@kdmnk

resources/views/dormitory/rooms/app.blade.php
sszajbely: Sadly I need the `fill ` CSS property to change not the `background-color`, so I couldn't use the materialize classes. But I tried my best to make the code readable in this regard.

kdmnk:
> > I also noticed some other things:
> > 
> > * RoomController - updateResidents:
> >   instead of:
> > 
> > ```
> >         foreach ($users as $user) {
> >             $user->update(['room' => null]);
> >         }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `$users->update(['room' => null]);` is enough
> 
> It doesn't seem to work for me. I get an error: `Method Illuminate\Database\Eloquent\Collection::update does not exist.` @kdmnk

Oh, yeah, you need this before get the users from db. So User::update(['room' => null]), and you do not even need the $users var.

app/Models/Room.php
```diff
@@ -20,12 +20,19 @@ class Room extends Model
 
     protected $withCount = ['users'];
 
+    public static $roomColors = [
+        3 => '#76ff03', //light-green accent-3
+        2 => '#ffee58', //yellow lighten-2
+        1 => '#ff9100', //orange accent-2
+        0 => '#e0e0e0', //grey lighten-2
```
kdmnk, line 27: I meant you do not need to use the # values. Pass 'light-green accent-3' as a css class.

kdmnk: And you do not need this attribute. Just return them in the color() function. Or do you use this elsewhere?

app/Models/Router.php
```diff
@@ -48,11 +48,17 @@ public function getFailStartDate()
 
     public function sendWarning()
     {
-        if ($this->failed_for == self::WARNING_THRESHOLD) {
+        if ($this->failed_for == self::WARNING_THRESHOLD && env('APP_DEBUG') == false) {
```
kdmnk: I meant do not increment the failed_for attribute if the app is in debug mode. I'm afraid  this way we still get emails after changing back to prod.

kdmnk (changes requested):
Okay, nice. Left some comments, otherwise, after that  can merge this. I'll do the livewire thing sometime.

app/Models/Room.php
sszajbely: I generate the legend using the attribute. And as I wrote, the color classes set the `background-color` style property and in svg I need to change the `fill `property.

sszajbely:
> > > I also noticed some other things:
> > > 
> > > * RoomController - updateResidents:
> > >   instead of:
> > > 
> > > ```
> > >         foreach ($users as $user) {
> > >             $user->update(['room' => null]);
> > >         }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > `$users->update(['room' => null]);` is enough
> > 
> > 
> > It doesn't seem to work for me. I get an error: `Method Illuminate\Database\Eloquent\Collection::update does not exist.` @kdmnk
> 
> Oh, yeah, you need this before get the users from db. So User::update(['room' => null]), and you do not even need the $users var.

`Non-static method Illuminate\Database\Eloquent\Model::update() cannot be called statically` I think I'll stay with the one that works :D

kdmnk (approved):

PR #142, opened 2022-10-07
Improve tenant handling

### LINKED ISSUE #127: Improve tenants handling
Tenants should only be considered "active" after they get approved and before their planned departure date.
Afterwards, hide them from the system, revoke their privileges and only allow them to request a renewal. 

Also, it could be good to store their room number, but just for quicker debugging in case they have internet problems (so this should be separate from the room assignment module)

sszajbely:
- The `tenant_until` date can maximally be 6 months in the future, this is also the date until which a tenant has internet.
- If a tenant's `tenant_until` date is in the past, they will be redirected to a page where they can update their `tenant_until` property.
- The `tenant_until` property of collegists with the tenant role will be updated automatically if their status is activated. The `tenant_until`'s new value will be 3 months after the end of the current semester.
- If an active collegist receives the tenant role, their `tenant_until` property will be updated in the same way.
- Because of the last two, active collegists won't have to set their planned departure date through the form.
- But now we can add the tenant role to former collegists living in the building and they will be able to update the planned departure (and get internet this way).
- Added an option for former tenants to become applicants. The `tenant_until` update form will show them this option if the application deadline is in the future and if the user is not a collegist.


app/Http/Controllers/Secretariat/RegistrationsController.php
```diff
@@ -44,8 +44,12 @@ public function accept(Request $request)
 
         $user->update(['verified' => true]);
         if ($user->hasRole(Role::TENANT)) {
-            $date = (Carbon::now()->addMonths(6)->gt($user->personalInformation->tenant_until.' 00:00:00')) ? ($user->personalInformation->tenant_until.' 00:00:00') : Carbon::now()->addMonths(6);
+            $date=Carbon::now()->addMonths(6);
+            if (Carbon::now()->addMonths(6)->gt($user->personalInformation->tenant_until.' 00:00:00')) {
+                $date = $user->personalInformation->tenant_until.' 00:00:00';
+            }
             $user->internetAccess()->update(['has_internet_until' => $date]);
+            $user->tenant_until->update(['tenant_until' => $date]);
```
kdmnk: ->personalInformation()->...

app/Http/Controllers/Secretariat/UserController.php
```diff
@@ -174,4 +175,75 @@ public function removeRole(Request $request, User $user, Role $role)
...
+                ->withInput();
+        }
+
+        $user = Auth::user();
+        $date=Carbon::now()->addMonths(6);
+        if (Carbon::now()->addMonths(6)->gt($request->tenant_until.' 00:00:00')) {
+            $date = $request->tenant_until.' 00:00:00';
+        }
+        $user->internetAccess()->update(['has_internet_until' => $date]);
+        $user->personalInformation->update(['tenant_until' => $date]);
```
kdmnk: personalInformation()

app/Http/Controllers/Secretariat/UserController.php
```diff
@@ -174,4 +176,67 @@ public function removeRole(Request $request, User $user, Role $role)
...
+
+        $user = Auth::user();
+        $date=Carbon::now()->addMonths(6);
+        $tenantUntilDate=Carbon::parse($request->tenant_until);
+        if (Carbon::now()->addMonths(6)->gt($tenantUntilDate)) {
+            $date = $tenantUntilDate;
+        }
+        $user->internetAccess()->update(['has_internet_until' => $date]);
+        $user->personalInformation()->update(['tenant_until' => $date]);
+
```
kdmnk: Modify the Usercontroller::updatePersonalInformation() accordingly and use that instead of this.

app/Http/Controllers/Secretariat/UserController.php
```diff
@@ -174,4 +176,67 @@ public function removeRole(Request $request, User $user, Role $role)
...
+     */
+    public function tenantToApplicant()
+    {
+        if (!Auth::user()->isTenant() || Auth::user()->isCollegist()) {
+            return abort(403);
+        }
+        $user = Auth::user();
+        $user->personalInformation()->update(['tenant_until' => null]);
+        $user->removeRole(Role::firstWhere('name', Role::TENANT));
+        $collegist = Role::firstWhere('name', Role::COLLEGIST);
```
kdmnk: delete var

app/Http/Middleware/PossibleRedirects.php
```diff
@@ -19,12 +19,19 @@ class PossibleRedirects
...
             && !$request->user()->hasActivated()) {
-            return redirect(route('secretariat.status-update.show'));
+                return redirect(route('secretariat.status-update.show'));
+            }
+            /** Active collegists living in
+            *   the dormitory as tenants are not affected as their tenant_until is set automatically until the end of the semester
+            */
+            if (!$request->is('users/tenant_update/*') && $request->user()->needsUpdateTenantUntil()) {
+                return redirect(route('users.tenant-update.show'));
+            }
```
kdmnk: Add more comments to this file (for every redirect: what happens when and why)

app/Models/User.php
```diff
@@ -411,6 +411,9 @@ public function addRole(Role $role, Workshop|RoleObject $object = null): bool
             if ($this->roles()->where('id', $role->id)->doesntExist()) {
                 $this->roles()->attach($role->id);
             }
+            if ($role->name == Role::TENANT && $this->isActive()) {
```
kdmnk: See below

app/Models/User.php
```diff
@@ -756,6 +791,9 @@ public function getStatus(): string
      */
     public function setStatusFor(Semester $semester, $status, $comment = null): User
     {
+        if ($semester==Semester::current() && $this->isTenant() && $status == SemesterStatus::ACTIVE) {
+            $this->setTenantUntilToSemesterEnd();
```
kdmnk: I'd add a setTenantDateForTenantCollegists() User model function that includes the checks also. Just call that everywhere after changing any status/role.

routes/web.php
```diff
@@ -71,6 +71,9 @@
     Route::post('/users/{user}/educational_information', [UserController::class, 'updateEducationalInformation'])->name('users.update.educational');
     Route::post('/users/{user}/roles/{role}', [UserController::class, 'addRole'])->name('users.roles.add');
     Route::delete('/users/{user}/roles/{role}', [UserController::class, 'removeRole'])->name('users.roles.delete');
+    Route::get('/users/tenant_update/show', [UserController::class, 'showTenantUpdate'])->name('users.tenant-update.show');
+    Route::post('/users/tenant_update/update', [UserController::class, 'updateTenantUntil'])->name('users.tenant-update.update');
```
kdmnk: delete this if you delete the function (see above)

kdmnk (changes requested):

kdmnk:
And also please review when the tenant until field is visible on the user page.

app/Http/Controllers/Secretariat/UserController.php
sszajbely: Can't it stay this way? I'll have to add another if in the redirect if I use another base url...

Other than that, if I modify the `updatePersonalInformation` function, we won't be able to mark everything required in the validator.

kdmnk: If the validator can not make this, then my first attempt would be to include the updatePersonalinformation blade in this page. I worked hard to have only one user update function :D

app/Models/User.php
```diff
@@ -795,6 +833,18 @@ public function verify($semester): User
...
     }
 
+    /**
+     * Sets the tenant_until date to the end of the current semester if the user is an active collegist with the tenant role.
+     * Added three months and two weeks to the end of the semester to allow the activation to happen.
+     */
+    public function setTenantDateForTenantCollegists(): bool
+    {
+        if ($this->isTenant() && $this->isActive()) {
+            return $this->personalInformation()->update(['tenant_until'=>Semester::current()->getEndDate()->addMonths(3)->addWeeks(2)]);
```
kdmnk, line 873: https://github.com/EotvosCollegium/mars/issues/144 a solution can be to set the date to next semester's start

app/Http/Controllers/Secretariat/UserController.php
```diff
@@ -174,4 +176,66 @@ public function removeRole(Request $request, User $user, Role $role)
...
+                ->withErrors($validator)
+                ->withInput();
+        }
+
+        $user = Auth::user();
+        $date=Carbon::now()->addMonths(6);
+        $tenantUntilDate=Carbon::parse($request->tenant_until);
+        if (Carbon::now()->addMonths(6)->gt($tenantUntilDate)) {
+            $date = $tenantUntilDate;
+        }
```
kdmnk: You can use Carbon's max function

resources/views/user/update_tenant_status.blade.php
```diff
@@ -0,0 +1,51 @@
...
+                        </blockquote>
+                        <x-input.datepicker
+                            id='tenant_until'
+                            required
+                            locale='user'
+                            :value="$user->personalInformation?->tenant_until" />
+                    </div>
+                    <div class="row">
+                        <x-input.button class="right red" text="general.save" />
+                    </div>
```
kdmnk: I meant that change the form here to an include to user.personalInformation. (You need to change the blade to only include the wanted fields for tenants: email, phone, tenant_until. Also, add a note in this file to update the tenant_until value in the form below).
And in the update controller function, now you only need to change the internet_until date also based on this.
Or do I miss something?

app/Http/Controllers/Secretariat/RegistrationsController.php
```diff
@@ -44,8 +44,13 @@ public function accept(Request $request)
 
         $user->update(['verified' => true]);
         if ($user->hasRole(Role::TENANT)) {
-            $date = (Carbon::now()->addMonths(6)->gt($user->personalInformation->tenant_until.' 00:00:00')) ? ($user->personalInformation->tenant_until.' 00:00:00') : Carbon::now()->addMonths(6);
+            $date=Carbon::now()->addMonths(6);
+            $tenantUntilDate=Carbon::parse($user->personalInformation->tenant_until);
+            if (Carbon::now()->addMonths(6)->gt($tenantUntilDate)) {
+                $date = $tenantUntilDate;
```
kdmnk: use min here also

kdmnk:
Please check if you can register properly (both as tenant/collegist). I got redirected, not sure why.

kdmnk (approved):
Merge at will

PR #178, opened 2022-10-30
Trying to fix mail to staff in FaultController

viktorcsimma:
Erika has complained she does not receive mails about faults. I see that the main difference between FaultController's and PrintController's mail management (the latter works) is that FaultController used "queue" instead of "send". May this be a problem? The [documentation](https://laravel.com/docs/9.x/mail) says it's about parallelization. Thank you in advance:)


kdmnk:
We usually use queue, I think that is not the problem. I tried this on the server (assigned staff role to myself) and I get the email. And you can always check the sent emails in our gmail account. I'd check the spam folders or sth similar. Also, if an email can not be sent, the user gets an 500 error.

viktorcsimma:
Hm… that means something’s probably wrong on Erika’s side. Thank you:)

PR #187, opened 2022-12-07
Setting printer size limit to 2 MB

viktorcsimma:
For some reason, it has problems with file sizes above ca. 2 MB. Maybe we should lower the official limit until we find out what the problem is. Thanks in advance:)


kdmnk:
I modified the value in .env on the server. This is just a fallback value ;)

PR #199, opened 2023-02-01
Voting system

### LINKED ISSUE #195: Adding anonymous vote system
Until the next General Assembly, I would like to add a feature through which people could vote using their Urán account. I already have a plan for the database which would not contain who voted for what; if you have any comments, I would appreciate them:)

Bildschirmfoto von 2023-01-17 18-36-00

viktorcsimma:
Addressing #195. Now it is essentially in a working state, but I'd be happy if you could check it:)


app/Models/Question.php
```diff
@@ -0,0 +1,83 @@
...
+    use HasFactory;
+
+    protected $fillable = ['title', 'sitting_id', 'max_options', 'opened_at', 'closed_at'];
+
+    public $timestamps = false;
+
+    public function sitting(): Sitting
+    {
+        return $this->belongsTo(Sitting::class)->first();
+    }
```
kdmnk:
```
    public function sitting(): BelongsTo
    {
        return $this->belongsTo(Sitting::class);
    }
```

kdmnk: This is the convention to define the relations. You can use `$question->sitting` with this to obtain the same thing, and also `$question->sitting()` for a query builder with which you can do many things,

app/Http/Controllers/StudentsCouncil/VotingController.php
```diff
@@ -0,0 +1,188 @@
...
+            'title' => $request->title,
+            'opened_at' => now(),
+        ]);
+
+        return view('student-council.voting.view_sitting', [
+            "sitting" => $sitting
+        ]);
+    }
+
+    public function viewSitting(Sitting $id)
```
kdmnk: Rename $id to $sitting. You just need to change it in the routes as well. Same below.

app/Models/Sitting.php
```diff
@@ -0,0 +1,78 @@
...
+    public function isClosed(): bool
+    {
+        return $this->closed_at!=null && $this->closed_at<=now();
+    }
+    public function open(): void
+    {
+        if ($this->isOpen() || $this->isClosed()) {
+            throw new Exception("tried to open sitting when it has already been opened");
+        }
+        $this->opened_at=now();
```
kdmnk:
```suggestion
        $this->update(['opened_at'=>now()]);
```

kdmnk: This way you don't need to call save() in the controllers, it saves automatically

kdmnk: Same for the others

app/Http/Controllers/StudentsCouncil/VotingController.php
```diff
@@ -0,0 +1,188 @@
...
+            function ($s) {
+                return trim($s);
+            },
+            array_filter(explode("\n", $request->options), function ($s) {
+                return !ctype_space($s);
+            })
+        );
+        if (count($options)==0) {
+            return back()->with('message', __('voting.at_least_one_option'));
+        }
```
kdmnk: With this return you will loose the already entered data in the input fields. If that can be a problem, you can add custom validation errors to the options field. You can see updatePersonalInformation function in the UserController for an example.

app/Http/Controllers/StudentsCouncil/VotingController.php
```diff
@@ -0,0 +1,188 @@
...
+        $validator = Validator::make($request->all(), [
+            'title' => 'required|string',
+            'max_options' => 'required|min:1'
+        ]);
+        $validator->validate();
+
+        $question = $id->addQuestion($request->title, $request->max_options, now());
+        foreach ($options as $option) {
+            $question->addOption($option);
+        }
```
kdmnk: Use native laravel functions instead: 
$id->questions()->create([...]);
and so on..

For this, you need to define the relations  in the models.

app/Http/Controllers/StudentsCouncil/VotingController.php
```diff
@@ -0,0 +1,188 @@
...
+            "question" => $id
+        ]);
+    }
+
+    public function saveVote(Question $id, Request $request)
+    {
+        $this->authorize('vote', $id); //this also checks whether the user has already voted
+
+        if ($id->max_options==1) {
+            $option=QuestionOption::where('id', $request->option)->first();
```
kdmnk: QuestionOption::find($request->option);
or
QuestionOption::findOrFail($request->option); if you are sure that it exists (the first one can return null)

app/Http/Controllers/StudentsCouncil/VotingController.php
```diff
@@ -0,0 +1,188 @@
...
+        $this->authorize('vote', $id);
+        return view('student-council.voting.vote', [
+            "question" => $id
+        ]);
+    }
+
+    public function saveVote(Question $id, Request $request)
+    {
+        $this->authorize('vote', $id); //this also checks whether the user has already voted
+
```
kdmnk, line 202: validation is missing

kdmnk: 'options' => 'array|max:'.$id->max_options', (now you dont need to check this below)
'options.*' => 'exists:...'

app/Models/Question.php
```diff
@@ -0,0 +1,83 @@
...
+        $this->closed_at=now();
+    }
+    public function addOption(string $title): QuestionOption
+    {
+        return QuestionOption::create([
+            'question_id' => $this->id,
+            'title' => $title,
+            'votes' => 0
+        ]);
+    }
```
kdmnk: delete and use relations

app/Models/Question.php
```diff
@@ -0,0 +1,83 @@
...
+    public function open(): void
+    {
+        if (!$this->sitting()->isOpen()) {
+            throw new Exception("tried to open question when sitting was not open");
+        }
+        if ($this->isOpen() || $this->isClosed()) {
+            throw new Exception("tried to open question when it has already been opened");
+        }
+        $this->opened_at=now();
```
kdmnk: use update

app/Models/Question.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+    public function close(): void
+    {
+        if ($this->isClosed()) {
+            throw new Exception("tried to close sitting when it has already been closed");
+        }
+        if (!$this->isOpen()) {
+            throw new Exception("tried to close sitting when it was not open");
+        }
+        $this->closed_at=now();
```
kdmnk: use update

app/Models/Question.php
```diff
@@ -0,0 +1,83 @@
...
+        ]);
+    }
+    public function isMultipleChoice(): bool
+    {
+        return $this->max_options>1;
+    }
+    public function hasVoted(User $user): bool
+    {
+        return DB::table('question_user')->where('question_id', $this->id)
+                ->where('user_id', $user->id)->exists();
```
kdmnk: Define a model for this table and use QuestionUser::where...
(even if the model won't do much things. Also maybe you should group these models in a directory. We need to start grouping the models eventually.

app/Models/Question.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+    public function canVote(User $user): bool
+    {
+        return $this->isOpen() && $user->isCollegist() && $user->isActive() && !$this->hasVoted($user);
+    }
+    public function getOptions()
+    {
+        return $this->hasMany(QuestionOption::class)->get();
+    }
```
kdmnk:
```suggestion
    public function options()
    {
        return $this->hasMany(QuestionOption::class)
    }
```

kdmnk: and move it to the beginning of the file

app/Models/QuestionOption.php
```diff
@@ -0,0 +1,40 @@
...
+    use HasFactory;
+
+    public $timestamps = false;
+
+    protected $fillable = ['question_id', 'title', 'votes'];
+
+    public function question(): Question
+    {
+        return $this->belongsTo(Question::class)->first();
+    }
```
kdmnk:
```suggestion
    public function question(): BelongsTo
    {
        return $this->belongsTo(Question::class);
    }
```

app/Models/QuestionOption.php
```diff
@@ -0,0 +1,40 @@
...
+
+    public function vote(User $user): void
+    {
+        $question=$this->question();
+        //if ($question->hasVoted($user)) throw new \Exception("user has already voted"); //had to take this out because of multi-option questions
+        if (!$question->isOpen()) {
+            throw new \Exception("question not open");
+        }
+        DB::table('question_user')->insert([
```
kdmnk: QuestionUser::create(...)

app/Models/QuestionOption.php
```diff
@@ -0,0 +1,40 @@
...
+
+    public function question(): Question
+    {
+        return $this->belongsTo(Question::class)->first();
+    }
+
+    public function vote(User $user): void
+    {
+        $question=$this->question();
+        //if ($question->hasVoted($user)) throw new \Exception("user has already voted"); //had to take this out because of multi-option questions
```
kdmnk: Maybe you should move this function to Questions model, providing a list of the chosen options.
Or worth considering moving this to the User model, as the user "votes".

app/Models/QuestionOption.php
```diff
@@ -0,0 +1,40 @@
...
+        if (!$question->isOpen()) {
+            throw new \Exception("question not open");
+        }
+        DB::table('question_user')->insert([
+            'question_id' => $question->id,
+            'user_id' => $user->id,
+            //'voted_at' => now() //now this is done by the automatic timestamps
+        ]);
+        $this->votes++;
+        $this->save();
```
kdmnk: $this->increment('votes');

app/Models/Sitting.php
```diff
@@ -0,0 +1,78 @@
...
+        if ($this->isClosed()) {
+            throw new Exception("tried to close sitting when it has already been closed");
+        }
+        if (!$this->isOpen()) {
+            throw new Exception("tried to close sitting when it was not open");
+        }
+        foreach ($this->questions() as $question) {
+            if ($question->isOpen()) {
+                $question->close();
+                $question->save();
```
kdmnk: delete save

app/Models/Sitting.php
```diff
@@ -0,0 +1,78 @@
...
+    public function close(): void
+    {
+        if ($this->isClosed()) {
+            throw new Exception("tried to close sitting when it has already been closed");
+        }
+        if (!$this->isOpen()) {
+            throw new Exception("tried to close sitting when it was not open");
+        }
+        foreach ($this->questions() as $question) {
+            if ($question->isOpen()) {
```
kdmnk: Does in makes sense to have yet to be opened questions for a closed sitting? If not, just close everything and delete the if.

app/Models/Sitting.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+    public function addQuestion(string $title, int $max_options=1, $opened_at=null, $closed_at=null): Question
+    {
+        return Question::create([
+            'sitting_id' => $this->id,
+            'title' => $title,
+            'max_options' => $max_options,
+            'opened_at' => $opened_at,
+            'closed_at' => $closed_at
+        ]);
```
kdmnk: Delete function, use relations. You can set the default values in the db migration

app/Models/Sitting.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+    public function questions()
+    {
+        return $this->hasMany(Question::class)->orderByDesc('opened_at')->get();
```
kdmnk:
```suggestion
        return $this->hasMany(Question::class)->orderByDesc('opened_at');
```

app/Policies/QuestionPolicy.php
```diff
@@ -0,0 +1,26 @@
...
+use App\Models\Question;
+use Illuminate\Auth\Access\HandlesAuthorization;
+
+class QuestionPolicy
+{
+    use HandlesAuthorization;
+
+    public function vote(User $user, Question $question): bool
+    {
+        return $question->canVote($user);
```
kdmnk: Inline the function content here. Delete the canVote function

app/Policies/SittingPolicy.php
```diff
@@ -0,0 +1,30 @@
...
+        return $user->isCollegist();
+    }
+
+    /**
+     * Determine whether the user can administer votings (add sitting, add or change question etc.).
+     */
+    public function administer(User $user)
+    {
+        return $user->hasRole(Role::SYS_ADMIN) || $user->hasRole([Role::STUDENT_COUNCIL => Role::PRESIDENT])
+            || $user->hasRole(Role::SECRETARY);
```
kdmnk: Please test this, but I think
```suggestion
        return $user->hasRole([Role::SYS_ADMIN, Role::STUDENT_COUNCIL => Role::PRESIDENT, Role::SECRETARY]);
```
should be the same.

database/migrations/2023_01_18_201909_create_voting_tables.php
```diff
@@ -0,0 +1,62 @@
...
+return new class () extends Migration {
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('sittings', function (Blueprint $table) {
+            $table->bigIncrements('id');
```
kdmnk:
```suggestion
            $table->id();
```

database/migrations/2023_01_18_201909_create_voting_tables.php
```diff
@@ -0,0 +1,62 @@
...
+            $table->string('title');
+            $table->datetime('opened_at');
+            $table->datetime('closed_at')->nullable();
+        });
+        Schema::create('questions', function (Blueprint $table) {
+            $table->bigIncrements('id');
```
kdmnk:
```suggestion
            $table->id();
```

database/migrations/2023_01_18_201909_create_voting_tables.php
```diff
@@ -0,0 +1,62 @@
...
+            $table->foreignId('sitting_id')
+                ->references('id')->on('sittings')
+                ->onDelete('cascade');
+            $table->string('title');
+            $table->integer('max_options')->default(1);
+            $table->datetime('opened_at')->nullable();
+            $table->datetime('closed_at')->nullable();
+        });
+        Schema::create('question_options', function (Blueprint $table) {
+            $table->bigIncrements('id');
```
kdmnk:
```suggestion
            $table->id();
```

database/migrations/2023_01_18_201909_create_voting_tables.php
```diff
@@ -0,0 +1,62 @@
...
+            $table->integer('votes')->default(0);
+        });
+        Schema::create('question_user', function (Blueprint $table) {
+            $table->foreignId('question_id')
+                ->references('id')->on('questions')
+                ->onDelete('cascade');
+            $table->foreignId('user_id')
+                ->references('id')->on('users')
+                ->onDelete('cascade');
+            $table->timestamps();
```
kdmnk, line 43: Not sure this works, but please try adding composite keys.
`$table->index(['question_id', 'user_id']);`
This should enforce in db level that the user can not vote for the same question twice. 
You can try this with inserting rows directly in mysql.

resources/views/student-council/voting/list.blade.php
```diff
@@ -0,0 +1,73 @@
+@extends('layouts.app')
+
+@section('title')
+<a href="#!" class="breadcrumb">@lang('voting.assembly')</a>
+@endsection
+@section('admin_module') active @endsection
```
kdmnk:
```suggestion
@section('student_council_module') active @endsection
```

kdmnk (changes requested):
Congratulations @viktorcsimma for your first bigger PR 

I commented a lot of things (read it through before start working on it), but these are mainly laravel conventions and readability improvements, the overall progress is nice.

I will take a look at the front end in the weekend.

Just call me if you have questions, it will be easier to answer.

app/Models/Question.php
sszajbely: https://laravel.com/docs/9.x/eloquent-relationships#main-content

database/migrations/2023_01_18_201909_create_voting_tables.php
sszajbely: Instead of `foreignId('user_id')` you should use `foreignIdFor(User::class)`. The result is probably the same, but it is less likely to make a mistake if you use it (and it is also easier to read).

resources/views/student-council/voting/new_question.blade.php
```diff
@@ -0,0 +1,37 @@
...
+                @csrf
+                <div class="card-content">
+                    <span class="card-title">@lang('voting.new_question')</span>
+                    <div class="row">
+                        <x-input.text s="12" type="text" text="{{ __('voting.question_title') }}" id="title" maxlength="100" required/>
+                    </div>
+                    <div class="row">
+                        {{-- TODO: this should be done in a less ugly way --}}
+                        <x-input.textarea id="options" s="12" l="10" text="{{ __('voting.options_instructions') }}" required/>
+                        <x-input.text type="number" min="1" max="3" value="1" s="12" l="2" id="max_options" text="{{ __('voting.max_options') }}" required/>
```
kdmnk: text="voting.options_instructions" will do it

resources/views/student-council/voting/list.blade.php
kdmnk: Same for the others

resources/views/student-council/voting/view_question.blade.php
```diff
@@ -0,0 +1,53 @@
...
+@section('content')
+
+<div class="row">
+    <div class="col s12">
+        <div class="card">
+            <div class="card-content">
+                <span class="card-title">{{ $question->title }}</span>
+                @cannot('view_results', $question)
+                <p>@lang('voting.only_after_closure')</p>
+                @endcan
```
kdmnk: <blockquote> instead of <p>

resources/views/student-council/voting/vote.blade.php
```diff
@@ -0,0 +1,52 @@
...
+                                <th>@lang('voting.options')</th>
+                            </tr>
+                        </thead>
+                        <tbody>
+                            @foreach($question->getOptions() as $option)
+                            <tr>
+                                <td>{{$option->title}}</td>
+                                <td>
+                                    @if($question->max_options==1)
+                                    <input style="opacity: 1; pointer-events: auto;" type="radio" name="option" value="{{ $option->id }}" text="{{ $option->title }}">
```
kdmnk: Do you need `style="opacity: 1; pointer-events: auto;"` here?

routes/web.php
```diff
@@ -209,4 +210,16 @@
...
+    Route::get('/voting/new_sitting', [VotingController::class, 'newSitting'])->name('voting.new_sitting');
+    Route::post('/voting/add_sitting', [VotingController::class, 'addSitting'])->name('voting.add_sitting');
+    Route::get('/voting/{id}', [VotingController::class, 'viewSitting'])->name('voting.view_sitting');
+    Route::post('/voting/{id}/close', [VotingController::class, 'closeSitting'])->name('voting.close_sitting');
+    Route::get('/voting/{id}/new_question', [VotingController::class, 'newQuestion'])->name('voting.new_question');
+    Route::post('/voting/{id}/add_question', [VotingController::class, 'addQuestion'])->name('voting.add_question');
+    Route::post('/voting/questions/{id}/close', [VotingController::class, 'closeQuestion'])->name('voting.close_question');
+    Route::get('/voting/questions/{id}/vote', [VotingController::class, 'vote'])->name('voting.vote');
+    Route::post('/voting/questions/{id}/vote', [VotingController::class, 'saveVote'])->name('voting.vote.save');
+    Route::get('/voting/questions/{id}', [VotingController::class, 'viewQuestion'])->name('voting.view_question');
```
kdmnk:
```suggestion
    Route::get('/', [SittingController::class, 'index'])->name('sittings.index');
    Route::get('/sittings/create', [SittingController::class, 'createSitting'])->name('sittings.create');
    Route::post('/sittings', [SittingController::class, 'storeSitting'])->name('sittings.store');
    Route::get('/sittings/{sitting}', [SittingController::class, 'showSitting'])->name('sittings.show');
    Route::put('/sittings/{sitting}/close', [SittingController::class, 'closeSitting'])->name('sittings.close');
    Route::get('/voting/questions/create', [VotingController::class, 'createQuestion'])->name('voting.questions.create');
    Route::post('/voting/questions', [VotingController::class, 'storeQuestion'])->name('voting.questions.store');
    Route::get('/voting/questions/{questions}', [VotingController::class, 'showQuestion'])->name('voting.questions.show');
    Route::put('/voting/questions/{question}/close', [VotingController::class, 'closeQuestion'])->name('voting.questions.close');
    Route::get('/voting/questions/{questions}/votes/create', [VotingController::class, 'createVote'])->name('voting.questions.votes.create');
    Route::post('/voting/questions/{questions}/votes', [VotingController::class, 'storeVote'])->name('voting.questions.votes.store');

```

kdmnk: Now here comes the fun part. Take a look at the CRUD (create-read-update-delete) concept here: https://laravel.com/docs/9.x/controllers#actions-handled-by-resource-controller
which gives us a nice structure for the whole resource handling. (We did not really use that in Mars yet)
If you want to follow this, you need to rename the links and controller functions (I suggest put it in a separate commit), but if you are in a hurry, don't worry, we will refactor later.

kdmnk (changes requested):
The UI also seems nice.
Please clean up the todos/comments and make sure that the tests are passing. If you have time, you can also write some new test cases but I know that we don't have good examples in this project, so feel free to leave that to me.

app/Http/Controllers/StudentsCouncil/VotingController.php
kdmnk: Maybe this one is ->questions()->attach([...])

app/Models/QuestionOption.php
viktorcsimma: That might be true... but it came across my mind that then we should also have to validate there whether the options really belong to the question; maybe that would be impractical. But on the other hand, making a BelongsTo query every time is impractical either, so that speaks for moving it to the questions.

app/Models/Sitting.php
viktorcsimma: Here it also checks whether it has already been closed; I think it would cause problems if we closed them two times.

database/migrations/2023_01_18_201909_create_voting_tables.php
viktorcsimma: @kdmnk here the problem is that now they do vote multiple times if there is a multiple-choice question. But maybe this could be solved if the vote() function were moved to the Question model, as you have suggested. (Probably you were right with that:D)

resources/views/student-council/voting/view_question.blade.php
viktorcsimma: Here I don't quite understand; have I done it wrong with the @lang directive?

resources/views/student-council/voting/vote.blade.php
viktorcsimma: Yes, it is now actually hacked together:D for some reason, it becomes invisible otherwise. I've tried with x-input but there it complained about unclosed ifs.

app/Models/QuestionOption.php
viktorcsimma: But then if we have to check there, that also means making a query each time... I don't know.

routes/web.php
viktorcsimma: Yes, I think I will take a look at it;)

app/Models/QuestionOption.php
kdmnk: I don't see now that it cannot be fixed, but maybe you are right. (Just to be clear, I wanted to move the vote function, but do whatever makes sense for you.) Validation should happen in the controller's validator anyways, maybe with a custom rule here as well.

database/migrations/2023_01_18_201909_create_voting_tables.php
kdmnk: I thought this table is just to keep track that the users voted for a specific question. You should not have multiple  question-user pair here even for multiple choice.

viktorcsimma: Hm... but then it really makes sense to move voting to the Question model; it would create a single question-user pair and then increment the votes. I think I will follow that route.

resources/views/student-council/voting/view_question.blade.php
kdmnk: Oh yeah, ignore this one please :D

routes/web.php
viktorcsimma: Done;) I removed the 'voting' prefixes from question-related routes; I think it is simpler for now.

kdmnk:
@viktorcsimma I tested the things, and made some minor improvements. About x-inputs: it's important that you need to end the tag with `/>`, maybe that was your problem. Now it should be good, also, we have a better input method instead of parsing a text along '\n'-s. That simplifies things. Take a look at the changes. I will make some tests and merge this asap, good work! 

One note: I found the sitting names pointless. We can make this in a way that also the committee-meetings can use the voting feature. This way, we can name the sittings "Közgyűlés" or "KommBiz ülés" etc. I leave this to you.

kdmnk (approved):
@viktorcsimma take a look at the changes and merge as you wish

kdmnk:
Also it might be nice if we list the people who voted with the results for a question.

kdmnk:
Another idea for the future: use one time codes that refreshes often to block out collegists voting from home. The code can be shown on the projection.

viktorcsimma:
Thank you very much for helping! I have also quickly added a passcode protection to questions; could you please check it out? I think it will be faster; I have only changed a few files.

kdmnk:
It's good for now, but if you want to refresh the password regularly, its a bit more tricky. Don't forget to create a password when you create a question.

kdmnk:
If you want to make the refresh, take a look at Jobs, Artisan commands (app/console folder), and the kernel's schedule function. But you also need to refresh the site. The most simple thing is probably to reload the page with JS every 5 seconds.

viktorcsimma:
Thank you! I'm not sure whether it will fit into the last week, so I think I will merge now; but maybe I will take a look at it;)

PR #213, opened 2023-03-01
Changed "modified_at" in PrintObserver to "now()"

### LINKED ISSUE #212: "Column 'modified_at' cannot be null" when printing
Some people have told us that printing throws a 500 error. I've checked the [logs](https://github.com/EotvosCollegium/mars/files/10865232/laravel.ERROR-2023-03-01.log) and found the corresponding file, but I don't quite understand. It complains about an INSERT and says:

```
[2023-03-01 08:53:43] production.ERROR: SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'modified_at' cannot be null (SQL: insert into `print_account_history` (`user_id`, `balance_change`, `free_page_change`, `deadline_change`, `modified_by`, `modified_at`) values (575, -12, 0, ?, 575, ?)) {"userId":575,"exception":"[object] (Illuminate\\Database\\QueryException(code: 23000): SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'modified_at' cannot be null (SQL: insert into `print_account_history` (`user_id`, `balance_change`, `free_page_change`, `deadline_change`, `modified_by`, `modified_at`) values (575, -12, 0, ?, 575, ?)) at /var/www/mars/vendor/laravel/framework/src/Illuminate/Database/Connection.php:759)
```

I've also checked the stacktrace and it lead to Printer.php:106. But it looks like this:

```php
private function printDocument()
{
    // Print file and return on error
    if (!$this->printFile()) {
        return back()->with('error', __('print.error_printing'));

    }

    // Update print account history
    $this->print_account->update(['last_modified_by' => Auth::user()->id]);
    foreach ($this->free_page_pool as $fp) {
        $fp['page']->update([
            'amount' => $fp['new_amount'],
            'last_modified_by' => Auth::user()->id
        ]);
    }

    // Update print account
    $this->print_account->decrement('balance', $this->cost);  //this is row 106

    return back()->with('message', __('print.success'));
}
```

So no inserts, only updates. What is even more puzzling is that it works perfectly when I try it. Do you have any ideas? Thank you very much in advance;)

viktorcsimma:
Tries to fix #212. Maybe there is already a null at the print account and it tries to write that into the history table.


app/Observers/PrintAccountObserver.php
```diff
@@ -32,7 +32,7 @@ public function updated(PrintAccount $printAccount)
                 'free_page_change' => 0,
                 'deadline_change' => null,
                 'modified_by' => $printAccount->last_modified_by,
-                'modified_at' => $printAccount->modified_at
+                'modified_at' => now()
```
kdmnk: Let's use $printAccount->modified_at ?? now()

kdmnk (changes requested):

app/Observers/PrintAccountObserver.php
kdmnk:
```suggestion
                'modified_at' => $printAccount->modified_at ?? now()
```

kdmnk (approved):

PR #243, opened 2023-04-02
Community Committee members can send the Epistola

### LINKED ISSUE #242: Add new role for Epistola editors
There were some members of the Communication Committee who would like to work on Epistola, but their current permissions aren't enough for that. I think we should create a new role for editing Epistola, besides the committee leader.

viktorcsimma:
Is to resolve #242.

But when I send a news item with this new role, I get this:
Bildschirmfoto von 2023-04-02 16-23-55
Is this a separate issue? Thank you in advance:)


sszajbely:
This error means that the EpistolaNews model doesn't have the `_token` as a fillable field. I don't know what the `$values` array is, but there is a `_token` key in it. If you need that value in the database you need to add it as a fillable field. If you don't need that, you will need to filter it out of the `$values` array.

kdmnk:
No, this error was due to the strict Laravel debug modes (btw the _token is the @csrf token, in these cases the models should be saved with $request->only(['attribute1', 'attribute2', ....])). I made that optionable now. While I'm here, I will redo this as we discussed, so we should just use the existing communication committee member/referent roles for that.

kdmnk (approved):

PR #275, opened 2023-06-07
Fixing can directive for the Sysadmin module in the navbar

viktorcsimma:
@AdamSulan has noticed he cannot reach the Sysadmin module; can it be because of this?

### DIFF
--- a/resources/views/layouts/navbar.blade.php
+++ b/resources/views/layouts/navbar.blade.php
@@ -108,7 +108,7 @@
                     </div>
                 </li>
                 @endcan
-                @can('is-collegist')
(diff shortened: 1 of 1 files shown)


kdmnk:
Well, not quite. We need the admin checkout to be visible for the collegists. We should somehove have collegist OR admin here then.

kdmnk:
I want to remove Gates in the future, they are a bit primitive. How about @if(user()->isAdmin() || user()->isCollegist()?

viktorcsimma:
That's a good idea:)

kdmnk (approved):

PR #311, opened 2023-07-27
Fix for application page throwing error on first try if user has no application form.

sszajbely:


app/Http/Controllers/Auth/ApplicationController.php
```diff
@@ -50,7 +50,7 @@ public function showApplicationForm(Request $request): View|RedirectResponse
             'faculties' => Faculty::all(),
             'deadline' => self::getApplicationDeadline(),
             'deadline_extended' => self::isDeadlineExtended(),
-            'user' => $request->user()
+            'user' => Auth::user(),
```
kdmnk:
```suggestion
            'user' => user(),
```

kdmnk (changes requested):

kdmnk (approved):

PR #329, opened 2023-08-13
Adding docker-compose.yml and information about running from terminal

viktorcsimma:
Based on instructions at TheBarbaraIsTaken/collegiumnostrum. This is the method I use now; so I though I would share it with you:)


kdmnk:
Thanks, I'll test if it does not break the current setup sometime during the week.

In the meantime, please rename every mention of collegiumnostrum.

kdmnk (changes requested):
Does not work for me, the db connection got refused. What DB config do you have in .env?

kdmnk:
If you want to work with this, add a new docker composer file (docker-compose.mysql.yml) and configure the prompts to use that one.

viktorcsimma:
The DB configuration in my .env is:
```
DB_CONNECTION=mysql
DB_HOST=mysql
DB_PORT=3306
DB_DATABASE=mars
DB_USERNAME=mars
DB_PASSWORD=secret
```
I don't know if this is helpful for you; have you used something different?

kdmnk (changes requested):
Okay, it works with your variables. Overwrite .env.example with those, please.
You can revert the last commit (`git revert 418b89e`)

viktorcsimma:
Done;)

PR #330, opened 2023-08-14
Same or unique rule

sszajbely:
Validation rule for determining if the given validation field is unique or the user already has it set.


kdmnk (approved):

kdmnk:
@machiato32 when can you fix the tests?

PR #335, opened 2023-09-11
Presence checks

### LINKED ISSUE #267: Store confirmed absence from general assemblies.
The president should be able to select a user and set a confirmed absence. This way the absence won't be calculated in the general assembly attendance requirements.

Additionally, the collegist should be able to see the confirmed absent status when they click on the assembly, on the semester evaluation page, and the semester evaluation export for the secretary. 

Try to have a database structure that can handle this and #293 easily.

### LINKED ISSUE #293: Make presence check for general assemblies
We should add separate questions for the presence check. Modify the current presence calculation so that the collegist is absent from the assembly if minimum 2 presence check failed.

The presence checks should be added and answered similarly as normal questions (with only one possible answer: I'm here), it only differs in the backend.

sszajbely:
- Created presence_checks table. This stores the presence checks during a general assembly.
- Created presence_check_user table. This is a many-to-many relationship for the users who signed in on a presence check.
  - Created all frontend and backend logic for these too.
- Created general_assembly_user table. This is a many-to-many relationship for the users who are excused from the general assembly.
  - Here the frontend and backend aren't ready.
- Users are counted as attending if 
  1. The number of presence checks is at least 3: they signed all but 2 presence checks.
  2. The number of presence cheks is less than 3: they singed all presence checks.
- Also refactored some code: moved Question related endpoints into a separate controller: GeneralAssemblyQuestionController. Like this the questions are treated as "children" of the general assembly, which makes the code cleaner.


app/Models/GeneralAssemblies/GeneralAssembly.php
```diff
@@ -133,11 +155,38 @@ public function close(): void
...
+    }
+
+    /**
+     * @return int The number of presence checks needed for a user to be counted as an attended user.
+     */
+    public function getPresenceChecksNeededAttribute(): int
+    {
+        $presenceCheckCount = $this->presenceChecks()->count();
+        return $presenceCheckCount <= 2 ? $presenceCheckCount : $presenceCheckCount - 2;
+    }
```
kdmnk, line 178: Why don't you use https://laravel.com/docs/10.x/eloquent-mutators#defining-an-accessor ?

database/migrations/2023_09_09_163542_create_presence_user_table.php
```diff
@@ -0,0 +1,34 @@
...
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('presence_check_user', function (Blueprint $table) {
+            $table->id();
+            $table->foreignIdFor(PresenceCheck::class)->constrained()->cascadeOnDelete();
+            $table->foreignIdFor(User::class)->constrained()->cascadeOnDelete();
+            $table->timestamps();
```
kdmnk: Create a composite key for user and presence check and use DB transactions if needed.

database/migrations/2023_09_10_194141_create_general_assembly_user_table.php
```diff
@@ -0,0 +1,34 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('general_assembly_user', function (Blueprint $table) {
+            $table->id();
+            $table->foreignIdFor(GeneralAssembly::class)->constrained()->cascadeOnDelete();
+            $table->foreignIdFor(User::class)->constrained()->cascadeOnDelete();
+            $table->timestamps();
+        });
```
kdmnk: is this for the excluded users? Clarify the name

resources/lang/en/voting.php
```diff
@@ -32,4 +32,11 @@
     'vote' => 'Vote',
     'voting' => 'Voting',
     'warning' => 'Warning: you cannot change your vote after you have submitted it.',
+    'presence_checks' => 'Presence checks',
+    'presence_check_title' => 'Presence check #:number',
+    'submit_presence_check' => 'I am here',
+    'close_presence_check' => 'Close presence check',
+    'presence_note' => 'Note',
+    'new_presence_check' => 'New presence check',
+    'excused_users' => 'Excused users',
```
kdmnk, line 41: We don't necessary need localisation here, but ok.

kdmnk (changes requested):

app/Models/GeneralAssemblies/GeneralAssembly.php
sszajbely: I've never used it before 😅

kdmnk (approved):
Fix the tests and let's give this a try.

PR #337, opened 2023-09-13
Additional improvements on the voting system.

### LINKED ISSUE #336: Allowing to open given assembly sittings at a later date
Currently, when we create a sitting, it automatically gets opened. It would be useful to create it the day before and fill it with questions we know beforehand.

viktorcsimma:
Main changes:
- Allows sittings and questions to be opened later (and thereby closes #336). The main advantage is that we can add questions to a not yet opened sitting; allowing us to prepare for the assembly the day before with questions we already know.
- Adds a page for displaying the passcode in (almost) full screen.
- Displays a message for people with no active status on the assembly page.

@kdmnk could we deploy this before the sitting? Or would that be too complicated? It's not essential, but would be helpful. Thanks in advance:)


viktorcsimma:
I've rebased it on @machiato32's commits, but also added a new feature, namely the fullscreen code (maybe I should have done that in a new commit). @kdmnk can you check it out?

kdmnk (approved):

PR #341, opened 2023-09-17
Displaying the private email of ethics commissioners instead of the university ones

viktorcsimma:
There has been a request from a commissioner to do this; and actually it's logical since we've done this with the president too. What do you think?


kdmnk (approved):
I'd argue that this is a good policy, but if they asked for it... Then, what happens with the next generation if they want to change it back? 
Tbh, I'd also hide the president's personal email.
Merge as you wish.

viktorcsimma:
What about adding a configuration option to `.env`?:D

viktorcsimma:
But I think I'll merge this for now. If someone complains, then we can still reevaluate the issue.

PR #342, opened 2023-09-23
Reviewing the algorithm for calculating workshop balances

viktorcsimma:
There was a strange phenomenon with the workshop balances: their sum was more than 49% of the amount received from kkt payments; in theory, it should be between 30% and 40%.

I've rewritten the algorithm calculating the data in the balances table; now it displays less people paying kkt and it seems to be consistent with the income numbers. @kdmnk could you check it out?

### DIFF
--- a/app/Http/Controllers/StudentsCouncil/EconomicController.php
+++ b/app/Http/Controllers/StudentsCouncil/EconomicController.php
@@ -70,6 +70,47 @@ class EconomicController extends Controller
         ]);
     }
 
+  /**
+    * Pay kkt and netreg to the receiver given.
+    * Also updates workshop balances
+    * and the internet access expiry date.
+    * Returns an array with the two transaction objects
+    * and the new expiry date.
+    *
+    * Used here and in the tests.
+    */
+   public static function payKKTNetregLogic(User $payer, User $receiver, int $kkt_amount, int $netreg_amount): array
+   {
+       // Creating transactions
+       $kkt = Transaction::create([
+           'checkout_id' => Checkout::studentsCouncil()->id,
+           'receiver_id' => $receiver->id,
+           'payer_id' => $payer->id,
+           'semester_id' => Semester::current()->id,
+           'amount' => $kkt_amount,
+           'payment_type_id' => PaymentType::kkt()->id,
+           'comment' => null,
+           'moved_to_checkout' => null,
+       ]);
+
+       $netreg = Transaction::create([
+           'checkout_id' => Checkout::admin()->id,
+           'receiver_id' => $receiver->id,
+           'payer_id' => $payer->id,
+           'semester_id' => Semester::current()->id,
+           'amount' => $netreg_amount,
+           'payment_type_id' => PaymentType::netreg()->id,
+           'comment' => null,
+           'moved_to_checkout' => null,
+       ]);
+
+       WorkshopBalance::generateBalances(Semester::current());
+
+       $new_expiry_date = \App\Http\Controllers\Network\InternetController::extendUsersInternetAccess($payer);
+
+       return [$kkt, $netreg, $new_expiry_date];
+   }
+
     /**
      * Pay kkt / netreg.
      */
@@ -86,7 +127,8 @@ class EconomicController extends Controller
 
         $payer = User::findOrFail($request->user_id);
         // the current user will be the receiver
-        [$kkt, $netreg, $new_internet_expire_date] = $payer->payKKTNetreg(Auth::user()->id, $request->kkt, $request->netreg);
+        [$kkt, $netreg, $new_internet_expire_date]
+            = self::payKKTNetregLogic($payer, Auth::user(), $request->kkt, $request->netreg);
 
         $internet_expiration_message = null;
         if ($new_internet_expire_date !== null) {
--- a/app/Http/Controllers/StudentsCouncil/EconomicController.php
+++ b/app/Http/Controllers/StudentsCouncil/EconomicController.php
@@ -85,7 +85,9 @@ class EconomicController extends Controller
         $validator->validate();
 
         $payer = User::findOrFail($request->user_id);
-        $payer->payKKTNetreg($request->kkt, $request->netreg);
+        $return_values = $payer->payKKTNetreg($request->kkt, $request->netreg);
+        $kkt = $return_values[0]; $netreg = $return_values[1];
+        $new_internet_expire_date = $return_values[2];
 
         $internet_expiration_message = null;
         if ($new_internet_expire_date !== null) {
@@ -94,7 +96,7 @@ class EconomicController extends Controller
             ]);
         }
 
-        Mail::to($payer)->queue(new \App\Mail\Transactions($payer->name, [$request->kkt, $request->netreg],
+        Mail::to($payer)->queue(new \App\Mail\Transactions($payer->name, [$kkt, $netreg],
                  "Tranzakció létrehozva", $internet_expiration_message));
 
         return redirect()->back()->with('message', __('general.successfully_added'));
--- a/app/Http/Controllers/StudentsCouncil/EconomicController.php
+++ b/app/Http/Controllers/StudentsCouncil/EconomicController.php
@@ -70,46 +70,46 @@ class EconomicController extends Controller
         ]);
     }
 
-  /**
-    * Pay kkt and netreg to the receiver given.
-    * Also updates workshop balances
-    * and the internet access expiry date.
-    * Returns an array with the two transaction objects
-    * and the new expiry date.
-    *
-    * Used here and in the tests.
-    */
-   public static function payKKTNetregLogic(User $payer, User $receiver, int $kkt_amount, int $netreg_amount): array
-   {
-       // Creating transactions
-       $kkt = Transaction::create([
-           'checkout_id' => Checkout::studentsCouncil()->id,
-           'receiver_id' => $receiver->id,
-           'payer_id' => $payer->id,
-           'semester_id' => Semester::current()->id,
-           'amount' => $kkt_amount,
-           'payment_type_id' => PaymentType::kkt()->id,
-           'comment' => null,
-           'moved_to_checkout' => null,
-       ]);
+    /**
+      * Pay kkt and netreg to the receiver given.
+      * Also updates workshop balances
+      * and the internet access expiry date.
+      * Returns an array with the two transaction objects
+      * and the new expiry date.
+      *
+      * Used here and in the tests.
+      */
+    public static function payKKTNetregLogic(User $payer, User $receiver, int $kkt_amount, int $netreg_amount): array
+    {
+        // Creating transactions
+        $kkt = Transaction::create([
+            'checkout_id' => Checkout::studentsCouncil()->id,
+            'receiver_id' => $receiver->id,
+            'payer_id' => $payer->id,
+            'semester_id' => Semester::current()->id,
+            'amount' => $kkt_amount,
+            'payment_type_id' => PaymentType::kkt()->id,
+            'comment' => null,
+            'moved_to_checkout' => null,
+        ]);
 
-       $netreg = Transaction::create([
-           'checkout_id' => Checkout::admin()->id,
-           'receiver_id' => $receiver->id,
-           'payer_id' => $payer->id,
-           'semester_id' => Semester::current()->id,
-           'amount' => $netreg_amount,
-           'payment_type_id' => PaymentType::netreg()->id,
-           'comment' => null,
-           'moved_to_checkout' => null,
-       ]);
+        $netreg = Transaction::create([
+            'checkout_id' => Checkout::admin()->id,
+            'receiver_id' => $receiver->id,
+            'payer_id' => $payer->id,
+            'semester_id' => Semester::current()->id,
+            'amount' => $netreg_amount,
+            'payment_type_id' => PaymentType::netreg()->id,
+            'comment' => null,
+            'moved_to_checkout' => null,
+        ]);
 
-       WorkshopBalance::generateBalances(Semester::current());
+        WorkshopBalance::generateBalances(Semester::current());
 
-       $new_expiry_date = \App\Http\Controllers\Network\InternetController::extendUsersInternetAccess($payer);
+        $new_expiry_date = \App\Http\Controllers\Network\InternetController::extendUsersInternetAccess($payer);
 
-       return [$kkt, $netreg, $new_expiry_date];
-   }
+        return [$kkt, $netreg, $new_expiry_date];
+    }
 
     /**
      * Pay kkt / netreg.
--- a/app/Http/Controllers/StudentsCouncil/EconomicController.php
+++ b/app/Http/Controllers/StudentsCouncil/EconomicController.php
@@ -86,7 +86,8 @@ class EconomicController extends Controller
 
         $payer = User::findOrFail($request->user_id);
         $return_values = $payer->payKKTNetreg($request->kkt, $request->netreg);
-        $kkt = $return_values[0]; $netreg = $return_values[1];
+        $kkt = $return_values[0];
+        $netreg = $return_values[1];
         $new_internet_expire_date = $return_values[2];
 
         $internet_expiration_message = null;
@@ -96,8 +97,12 @@ class EconomicController extends Controller
             ]);
         }
 
-        Mail::to($payer)->queue(new \App\Mail\Transactions($payer->name, [$kkt, $netreg],
-                 "Tranzakció létrehozva", $internet_expiration_message));
+        Mail::to($payer)->queue(new \App\Mail\Transactions(
+            $payer->name,
+            [$kkt, $netreg],
+            "Tranzakció létrehozva",
+            $internet_expiration_message
+        ));
 
         return redirect()->back()->with('message', __('general.successfully_added'));
     }
--- a/app/Http/Controllers/StudentsCouncil/EconomicController.php
+++ b/app/Http/Controllers/StudentsCouncil/EconomicController.php
@@ -85,33 +85,8 @@ class EconomicController extends Controller
         $validator->validate();
 
         $payer = User::findOrFail($request->user_id);
+        $payer->payKKTNetreg($request->kkt, $request->netreg);
 
-        // Creating transactions
-        $kkt = Transaction::create([
-            'checkout_id' => Checkout::studentsCouncil()->id,
-            'receiver_id' => user()->id,
-            'payer_id' => $payer->id,
-            'semester_id' => Semester::current()->id,
-            'amount' => $request->kkt,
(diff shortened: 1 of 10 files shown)


kdmnk:
I see, yet I still do not understand where the difference is coming from.
You changed two things:
- Check if the user paid: Instead of checking if the user is not returned by the hasToPayKKTNetregInSemester scope (which returns all active collegists who don't have either kkt or netreg transaction in the semester), you check if the individual user has any kkt transaction (which is more database calls, but that's another problem). Since active memberships are checked anyway before, this change should not result in any difference (given that everyone who has netreg transaction also has kkt transaction, and the contrary). 
- Use the paid amount instead of the amount in the config. I see the point, but do we want to have a workshop where they get less amount only because they have some people in the student council who pay 0 ft?  My opinion is that the workshop balance should be calculated based solely on the number of members, and not who paid how much. 

How did you check and validate the new numbers?

If we want to fix this in the long term, write test cases. This module is one of the last crucial parts where we don't have tests yet.
Let me know what you think.

viktorcsimma:
For the amount part, I agree with you; I've changed it back now. For the other part, I'm thinking about how to solve that. Now I've left it, and it seems to work; but if I understand well, the problem with those paying 0 Ft persists. (Or might that be the reason for the difference?)

viktorcsimma:
Now I've changed it to nullable; the numbers are still OK, but if I understand right, there aren't any transactions recorded for student council members, so the problem would still persist. Is that correct?

I've actually tested the code on the live site; it was consistent with the real values, but I still don't know the reason. Maybe I'll take a look at `hasToPayKKTNetregInSemester`.

kdmnk:
I insist on creating tests for this. There are multiple examples now in the project. That should also reveal the answer to our questions.

viktorcsimma:
Have you thought of something like this? I would write more, of course.

Actually, after some seeding, this fails for both the old and the new version failed... There, the sum of workshop balances was 10200 in the old version and 6600 in the new version; while the income from kkt was always 24000. So both of them was too little. I will check this out.

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,49 @@
...
+        echo(config('custom.workshop_balance_resident'));
+        echo("\n");
+        echo(config('custom.workshop_balance_extern'));
+        echo("\n");
+        */
+
+        $diff1 = config('custom.workshop_balance_resident') * $skkt - $sum;
+        $diff2 = config('custom.workshop_balance_extern') * $skkt - $sum;
+        $this->assertTrue($diff1 * $diff2 <= 0);
+    }
```
kdmnk: Aren't all the values 0 here?

You should test if after someone pays, the workshop balance for the user got modified with the correct amount.
Test externs, residents, with single and multiple workshops.

You can keep this test case but this is not so useful on its own.

kdmnk (changes requested):

viktorcsimma:
Now I think I've written something like that:) @kdmnk could you check it out?

Maybe a feature test for a complete HTTP request would also be useful; but that could be done in a separate PR, if you agree.

viktorcsimma:
One more thing: I had to rewrite `isResident()` and `isExtern()` because I got an error that there is a string somewhere where a model object should be. Do you have an idea what could have caused this?

viktorcsimma:
And maybe there could be another case where the test user has multiple workshops.

app/Http/Controllers/StudentsCouncil/EconomicController.php
```diff
@@ -85,41 +85,24 @@ public function payKKTNetreg(Request $request)
         $validator->validate();
 
         $payer = User::findOrFail($request->user_id);
+        $return_values = $payer->payKKTNetreg($request->kkt, $request->netreg);
```
kdmnk: [$kkt, $netreg, $new_internet_expire_date] = $payer->payKKTNetreg(...)

app/Models/User.php
```diff
@@ -681,13 +681,10 @@ public function setCollegist($objectName): void
...
-            return $this->roles()
-            ->where('role_id', Role::collegist()->id)
-            ->where('object_id', RoleObject::firstWhere('name', Role::RESIDENT)->id)
-            ->exists();
-        }
-        return $this->hasRole([Role::COLLEGIST => Role::RESIDENT]);
+        return $this->roles()
+        ->where('role_id', Role::collegist()->id)
+        ->where('object_id', RoleObject::firstWhere('name', Role::RESIDENT)->id)
+        ->exists();
```
kdmnk: Just comment the previous one out and add a todo. 
We need to remove the first branch when we detach roles from the applicants so it should just be  $this->hasRole(...).

For the bug you see, idk, we should find it asap, but if you can not you can leave this like that.

app/Models/User.php
```diff
@@ -869,25 +863,64 @@ public function sumOfActiveFreePages(): int
...
+     * and the internet access expiry date.
+     * Returns an array with the two transaction objects
+     * and the new expiry date.
+     */
+    public function payKKTNetreg(int $kkt_amount, int $netreg_amount)
+    {
+        // Creating transactions
+        $kkt = Transaction::create([
+            'checkout_id' => Checkout::studentsCouncil()->id,
+            'receiver_id' => Checkout::studentsCouncil()->handler_id,
```
kdmnk: the receiver can be any collegist that the student council delegates to, not necessarily the handler. In both cases.

kdmnk: Despite how they do this nowadays.

app/Models/User.php
```diff
@@ -869,25 +863,64 @@ public function sumOfActiveFreePages(): int
...
+    }
+
+    /**
+     * Returns the payed kkt amount in the semester; or null if the user has not payed kkt.
      * @param Semester $semester
-     * @return int
+     * @return ?int
      */
-    public function payedKKTInSemester(Semester $semester): int
+    public function payedKKTInSemester(Semester $semester): ?int
```
kdmnk: Would you rename everything to "paidSomething" while you are here, please?

app/Models/Workshop.php
```diff
@@ -69,6 +70,23 @@ public function externs()
...
+        return $this->hasMany(WorkshopBalance::class);
+    }
+
+    /**
+     * Returns the balance for a given semester
+     * (by default the current one).
+     */
+    public function balance(int $semester = null): WorkshopBalance
+    {
+        return $this->balances()->where('semester_id', $semester ?? Semester::current()->id)->firstOrFail();
```
kdmnk: Won't this die if we switch to next semester and the current workshop balances are not generated yet?
I'd allow null value and handle that.

app/Models/WorkshopBalance.php
```diff
@@ -58,7 +65,7 @@ public static function generateBalances($semester_id)
             $not_yet_paid = 0;
             foreach ($workshop->users as $member) {
                 if (isset($active_users[$member->id])) {
-                    if (!isset($users_has_to_pay_kktnetreg[$member->id])) {
+                    if (!is_null($member->payedKKTInSemester(Semester::find($semester_id)))) {
```
kdmnk: Get the semester outside of the loop. Or even outside this function and get the semester model as param. It makes our type checking stronger.

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Tests how workshop balances change
+     * when a user pays kkt.
+     * Does it both for externs and residents.
+     */
+    public function testPayment()
+    {
+        foreach ([true, false] as $is_extern) {
```
kdmnk: nice trick, but maybe cleaner to put these into two separate test cases.

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,104 @@
...
+            $workshop_id = rand(1, count(Workshop::ALL));
+            $workshop = Workshop::findOrFail($workshop_id);
```
kdmnk: You can use Workshop::all()->random() and then get the id from that.

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,104 @@
...
+            $user = User::factory()->create();
+
+            if ($is_extern) {
+                $user->setExtern();
+            } else {
+                $user->setCollegist(\App\Models\Role::RESIDENT);
```
kdmnk: ->setResident?

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,104 @@
...
+            }
+
+            $user->setStatusFor(Semester::current(), SemesterStatus::ACTIVE);  // important!
+
+            $user->workshops()->attach($workshop_id);
```
kdmnk: FYI I believe you can simply pass the workshop, but whatever

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,104 @@
...
+
+            $with_same_status =
+                $is_extern ? $workshop->externs() : $workshop->residents();
+            $this->assertTrue(
+                $with_same_status->filter(function (User $u) use ($user) {return $u->id == $user->id;})->count()
+                  == 1
+            );
```
kdmnk: What's the point here? this only checks that you set previously and that the user does not appear twice in the collection which can never be the case, can it?

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,104 @@
...
+            $user->payKKTNetreg(config('custom.kkt'), config('custom.netreg'));
+            $balance->refresh();
+            $new = $balance->allocated_balance;
+
+            $this->assertTrue($user->payedKKT() == config('custom.kkt'));
+
+            $this->assertTrue($new - $old == round(
+                ($is_extern ? config('custom.workshop_balance_extern')
+                                                           : config('custom.workshop_balance_resident'))
+                                               * config('custom.kkt')
```
kdmnk: Use precoded values instead of config values to make this easier to read and debug. You can have another test case that the pay function uses the config values accordingly.

kdmnk (changes requested):
Okay looking better. Please take a look at my comments. And yes, please write a case to test multiple workshops, that's a common case we have. So I imagine 4 cases, combining resident/extern and one/multiple workshops. 
Have you found the original bug yet?

An HTTP test is not necessarily here as we test those anyways during development. Those are useful for refactorings though, but we most importantly lack unit tests to test different edge cases like these.

tests/Unit/EconomicTest.php
viktorcsimma: The point was that I wanted to know that the status setting occurred successfully. But I agree it does not belong here.

app/Models/User.php
viktorcsimma: I now added it as a parameter.

app/Models/User.php
viktorcsimma: I don't understand this here; should I generalise the function to be able to pay anything? Now it is specially about kkt.

app/Models/User.php
viktorcsimma: Oh; it was the cache:D After running `php artisan optimize:clear`, it was fine. I revert it.

tests/Unit/EconomicTest.php
viktorcsimma: I had a problem with the ratios, or rather a design question: now `generateBalances` itself uses the config values for them; should we provide them as parameters or store them as properties of the semester? Or simply leave this as it is?

viktorcsimma:
Now I've added new test cases; I've also removed the first one (there was some problem with it, and you have mentioned it's not useful anyway).

Can you check out these and my questions, too? Sometime when it's convenient for you – until then, happy holidays;)

app/Http/Controllers/StudentsCouncil/EconomicController.php
```diff
@@ -85,41 +85,22 @@ public function payKKTNetreg(Request $request)
         $validator->validate();
 
         $payer = User::findOrFail($request->user_id);
+        // the current user will be the receiver
+        [$kkt, $netreg, $new_internet_expire_date] = $payer->payKKTNetreg(Auth::user()->id, $request->kkt, $request->netreg);
```
kdmnk: As a new thought, this adds another method to the User model which we want to keep as simple as possible.
I was thinking a lot about where would this be placed the best, in the Transaction model or somehow in CheckoutHandler, but these actions are typical controller actions, so you don't need to extract them away. We can make the function smaller with automatic email sending with observers, but that would mean refactoring the whole CheckoutHandler.
Just revert this change for now, sorry :D

app/Models/User.php
kdmnk: Just rename it to make it gramatically correct, paidKKT and paidNetreg.

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,114 @@
...
+        } else {
+            $user->setResident();
+        }
+
+        $user->setStatusFor(Semester::current(), SemesterStatus::ACTIVE);  // important!
+        foreach($workshops as $workshop) {
+            $user->workshops()->attach($workshop);
+        }
+
+        $this->assertTrue(is_null($user->payedKKT()));
```
kdmnk: unnecessary

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,114 @@
...
+        WorkshopBalance::generateBalances(Semester::current()); // this also ensures the balance won't be null
+
+        $balances = array_map(function ($workshop) {return $workshop->balance();}, $workshops);
```
kdmnk: use collection methods. 
$old_balances = $workshops->map(fn ($workshop) => $workshop->balance()->allocated_balance). 
$new_balances = same as above, i don't even think you need to refresh the vars this way.

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,114 @@
...
+        // we give the payer's id as the receiver's id
+        // because Checkout::studentsCouncil()->handler_id returns null
+        $user->payKKTNetreg($user->id, self::TEST_KKT, self::TEST_NETREG);
+
+        $this->assertTrue($user->payedKKT() == self::TEST_KKT);
+
+        $news = array_map(function ($balance) {return $balance->fresh()->allocated_balance;}, $balances);
+
+        $amount_for_workshops =
+            ($is_extern ? config('custom.workshop_balance_extern')
```
kdmnk: remove these configs as well. This also ensures that the pipelines test with the same data.

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,114 @@
...
+                        : config('custom.workshop_balance_resident'))
+                * self::TEST_KKT;
+        for ($i = 0; $i < $number_of_workshops; ++$i) {
+            // to avoid floating point equality checking
+            // and to allow some error from rounding:
+            $this->assertTrue(abs(
+                $number_of_workshops * ($news[$i] - $olds[$i])
+                                - $amount_for_workshops
+            ) < self::THRESHOLD_DIFF);
+        }
```
kdmnk: This is not readable. Use assertEqualsWithDelta and avoid for-loop indexes. Loop through the workshops, use dicts or whatever. Even easier to check if the workshop balances equal to an expected hard-coded value that you calculated before, no need to calculate the diffs programatically. Trust me it's better for debugging tests.

You should be able to set the threshold below 0.1. If not, something's wrong. If we round, still, the sum should all add up to the originally paid amount. Maybe you need to modify the logic behind it.

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,114 @@
...
+        $workshops = Workshop::all()->random($number_of_workshops)->all();
+        $user = User::factory()->create();
+
+        if ($is_extern) {
+            $user->setExtern();
+        } else {
+            $user->setResident();
+        }
+
+        $user->setStatusFor(Semester::current(), SemesterStatus::ACTIVE);  // important!
```
kdmnk: ->setStatus(SemesterStatus::ACTIVE);

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,114 @@
...
+        foreach($workshops as $workshop) {
+            $user->workshops()->attach($workshop);
+        }
```
kdmnk: $user->workshops()->sync(Workshop::all()->random(2))

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,114 @@
...
+    private function testPayment(bool $is_extern, int $number_of_workshops = 1): void
+    {
+        $workshops = Workshop::all()->random($number_of_workshops)->all();
+        $user = User::factory()->create();
+
+        if ($is_extern) {
+            $user->setExtern();
+        } else {
+            $user->setResident();
+        }
```
kdmnk: $user->setCollegist($is_extern ? Role::EXTERN : Role::RESIDENT)

tests/Unit/EconomicTest.php
```diff
@@ -0,0 +1,114 @@
...
+    {
+        $this->testPayment(false);
+    }
+
+    /**
+     * The concrete call of testPayment for externs with two workshops.
+     */
+    public function testExternPaymentWithTwoWorkshops()
+    {
+        $this->testPayment(true, 2);
```
kdmnk: These are also not readable. How about passing the user as a parameter?
I suggested some changes to make the user creation logic smaller above. 

I might also move the assertion logic here with the hard-coded amounts described above.
And with that, you probably won't even need these functions :D

kdmnk (changes requested):
These comments are just to make the code more readable. Good job

kdmnk:
I'm not saying that it's perfect, but you can take a look at these tests, here the assertions are clear what they tests and how they got calculated (eg. hard coded  `400 * (100 - (100 - 25) / 2)`):
https://github.com/DevsWithDodo/dodo_api/blob/master/tests/Feature/PurchaseCreateTest.php

app/Models/User.php
viktorcsimma: Oh, I see:)

app/Http/Controllers/StudentsCouncil/EconomicController.php
viktorcsimma: What about writing a static method in the EconomicController class? That was where this code was originally written. I would rather write a method for it somewhere, so that we don't need to duplicate the code in the test. See the code I'm going to push soon.

tests/Unit/EconomicTest.php
viktorcsimma: Passing the user is a good idea; that way, it's more evident what we are doing while we also avoid much of code duplication. I think I'll go for that:)

kdmnk (approved):

kdmnk:
Sorry about the hassle of such a little feature, but I think it looks really good and clean now. Merge as you wish

viktorcsimma:
It's fine; thank you for your help:) eventually, I'll have to know how to sustain this code quality.

PR #346, opened 2023-10-02
Allowing externs to be assigned to rooms

### LINKED ISSUE #345: Assigning resident externs to rooms
There have been complaints that resident externs cannot be added to rooms, because their status is officially Extern and only residents are allowed.

My first idea is simply to lift this restriction. Another solution is to create a new status for resident externs, but as far as I remember, @machiato32 did not want to formalise it (I don't know if it's still actual). What do you think?

viktorcsimma:
Closes #345. Simply allows all active collegists (including externs) to be assigned to rooms.


kdmnk (changes requested):
Filter on collegists at least

viktorcsimma:
As I remember, only collegists can have an active status; am I wrong? I have added an additional filter anyway.

kdmnk:
True, but there is no connection in them in the database. For example, after users become alumni, they still have statuses for the old semesters. (not relevant in this case, but still)

kdmnk (approved):

PR #354, opened 2023-10-23
New fields for profile picture, research topics and extra information in profile

### LINKED ISSUE #349: Adding new fields to users' profile (research topic, photo etc.)
The student council has requested workshop secretaries to provide more information about members. This includes:
- Their research topic.
- How occupied they are.
- A photo of them.
- _(Maybe some other personal data; I will have to ask about that.)_

The secretaries have suggested this could be added to Mars. What do you think?

viktorcsimma:
Closes #349. Now, there are two extra fields; with one for research topics (which could be included in the end-of-semester form) and one for everything else requested by secretaries and the student council. The profile picture is also displayed and can be updated. What do you think?


kdmnk:
It's fine. We will refine the user policies later anyway.
Display the profile picture while you are working on this (copy from the applications view).

kdmnk:
No, sorry, add it to the educational information, not to the personal one.

viktorcsimma:
Now I have added a field for the profile picture and done the localization. So it could even be merged; only the layout of the profile picture is what I'm unsure about. Can you check it out?

kdmnk (changes requested):
Nice, thanks. 
For the UI to make it better:
There's a card in the application page under the files tab. Generalise that and put that into the views/user.
Delete the profile picture logic from the application controller. Make a new endpoint to update the profile picture that points to a fn in the UserController and use that endpoint in the generalised component.
Add that card on the profile page before the personal information card.

It should then handle the profile page uploads from the application page and from the profile page. Maybe change the title from "Profile picture" to the person's name in case of the profile page.

viktorcsimma:
Done:)

kdmnk (approved):
While I was trying out, I found some unused stuff ($only_tenant_until) so I decided to push a cleanup here.
Also, the collapsible card was buggy for me, so I just made it constant while I was there and simplified the logic.
Note that you can use @method('delete')  to denote a deleting form.
Now its fine by me, but if you have time, could you test the mobile view? I'm not sure if that works properly. Or leave that to later, that's fine.

kdmnk:
Maybe one more thing: Hide the new educational information fields from the applications.

viktorcsimma:
I've tried to improve the mobile interface; a problem that has remained is the too large gap between the two buttons if they get below each other. I've also fixed something in the background logic.

I think I'll merge this for now. Thanks for your help:)

PR #356, opened 2023-10-28
Adding receipts to expenses

### LINKED ISSUE #331: Add receipts to payments
Users should be able to (or required to) attach photos of the receipts when they upload an expense

viktorcsimma:
Closes #331. Receipts are required for expenses; they can also be downloaded from the list, but they cannot be modified.

### DIFF
--- a/app/Utils/CheckoutHandler.php
+++ b/app/Utils/CheckoutHandler.php
@@ -15,6 +15,7 @@ use Illuminate\Support\Collection;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Mail;
+use Illuminate\Support\Facades\Storage;
 use Illuminate\Support\Facades\Validator;
 use Illuminate\Validation\ValidationException;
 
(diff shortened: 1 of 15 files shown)


resources/views/utils/checkout/add-expense.blade.php
```diff
@@ -33,6 +33,8 @@
                 <x-input.select m=6 l=6 id="payer" text="Fizető" :elements="\App\Models\User::collegists()" default="{{user()->id}}" :formatter="function($user) { return $user->uniqueName; }" />
                 <x-input.checkbox m=6 l=6 id="paid" checked text="A tartozás kifizetésre került"/>
                 @endcan
+                <x-input.file s="12" xl="8" id="receipt" style="margin-top:auto" accept=".pdf,.jpg,.png,.jpeg,.gif"
```
kdmnk: Only one question: why xl=8? I'd just delete it.

app/Utils/CheckoutHandler.php
```diff
@@ -209,6 +212,9 @@ public function addExpense(Request $request): RedirectResponse
             'paid_at'           => $paid ? Carbon::now() : null,
         ]);
 
+        $path = $request->file('receipt')->store('receipts');
```
kdmnk: Wrap these and the Transaction::create in a DB::transaction(...), so if either of the logic fails it backtracks. Same with deletion.

kdmnk (changes requested):

viktorcsimma:
Done:)

kdmnk (approved):
Codewise it's fine, you can test it live :P

One comment though: you can put the mail sending into the transaction as well. Now if that fails, the user gets an error message, yet the transactions are created properly. Depends how we want it.

viktorcsimma:
Hm... I have arguments both for and against it. Such a situation would be confusing; but then if anything happens with the mail function, it does not paralyse the function itself.

I think I'm going to merge it now; thank you:)

Issue #291, opened 2023-07-04
Refine guest roles/permissions


kdmnk:
Only handle users as guests if they are guests currently.
Consider deleting the guest role or detach them automatically.

kdmnk:
Part of #392, if relevant at all.

PR #358, opened 2023-10-28
Adding a resident-extern role

### LINKED ISSUE #347: Creating a resident extern status
It seems that the resident extern status is going to stay in the coming years, and it would be useful to track it. I quote @szentep:

_"Sziasztok! A mai igazgatói beszélgetésen világossá vált, hogy a BB státusz marad. Az utóbbi években jó megoldásnak bizonyult a helyek feltöltésére, hiszen ezek száma a felvételi végéig nem végleges, sőt, még szeptemberben is változik. Szerintem az Uránban is érdemes lenne bevezetni"_

As I remember, @machiato32 has said it would not be desirable as it would formalise something that should be abolished soon. But now, it looks like it's not going to be the case.

What do you think?

viktorcsimma:
Closes #347. It's largely done; there might be some bugs that I haven't noticed.

### DIFF
--- a/app/Http/Controllers/Secretariat/SemesterEvaluationController.php
+++ b/app/Http/Controllers/Secretariat/SemesterEvaluationController.php
@@ -157,7 +157,7 @@ class SemesterEvaluationController extends Controller
                         return back()->with('error', "A státusz megadása kötelező!")->with('section', $request->section);
                     }
                     $user->setStatusFor(Semester::next(), $request->next_status, $request->next_status_note);
-                    if ($request->has('resign_residency') && $user->isResident()) {
+                    if ($request->has('resign_residency') && $user->resides()) {
                         $user->setExtern();
                     }
                 }
--- a/app/Models/ApplicationForm.php
+++ b/app/Models/ApplicationForm.php
@@ -119,6 +119,9 @@ class ApplicationForm extends Model
         return $this->belongsTo('App\Models\User')->withoutGlobalScope('verified');
     }
 
+    /**
+     * Lists the files belonging to the application form.
+     */
     public function files(): \Illuminate\Database\Eloquent\Relations\HasMany
     {
         return $this->hasMany('App\Models\File');
--- a/app/Models/ApplicationForm.php
+++ b/app/Models/ApplicationForm.php
@@ -286,7 +286,7 @@ class ApplicationForm extends Model
             $missingData[] =  'Megjelölt kar';
         }
 
-        if (!$user->isResident() && !$user->isExtern()) {
+        if (!$user->resides() && !$user->isExtern()) {
             $missingData[] =  'Megjelölt collegista státusz';
         }
 
--- a/app/Exports/UsersSheets/CollegistsExport.php
+++ b/app/Exports/UsersSheets/CollegistsExport.php
@@ -67,7 +67,8 @@ class CollegistsExport implements FromCollection, WithTitle, WithMapping, WithHe
         return [
             '=HYPERLINK("'.route('users.show', ['user' => $user->id]).'", "'.$user->name.'")',
             $user->educationalInformation?->neptun,
-            $user->isResident() ? 'Bentlakó' : ($user->isExtern() ? 'Bejáró' : ($user->isAlumni() ? "Alumni" : ($user->isTenant() ? "Vendég" : ""))),
+            $user->isResident() ? 'Bentlakó' : ($user->isResidentExtern() ? 'Bentlakó-bejáró' :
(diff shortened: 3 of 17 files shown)


viktorcsimma:
What should we do? I really think this would be useful; now that this is going to be the new status quo. @machiato32?

sszajbely (commented):
The changes are alright. If this status really needs to stay, I guess I'll have to approve it. What do you think @kdmnk?

sszajbely:
Also, please fix the merge conflicts @viktorcsimma

viktorcsimma:
Something is not OK yet; it crashes at the economic committee page. I'll check it out soon.

sszajbely:
Ah, sorry didn't test it, just looked at the code.

app/Http/Controllers/Secretariat/SemesterEvaluationController.php
```diff
@@ -157,7 +157,7 @@ public function store(Request $request)
                         return back()->with('error', "A státusz megadása kötelező!")->with('section', $request->section);
                     }
                     $user->setStatusFor(Semester::next(), $request->next_status, $request->next_status_note);
-                    if ($request->has('resign_residency') && $user->isResident()) {
+                    if ($request->has('resign_residency') && $user->resides()) {
```
kdmnk, line 160: resides() is a scope, which I think returns a query builder (test it in tinker). Check if these checks work as expected.

app/Models/ApplicationForm.php
```diff
@@ -286,7 +289,7 @@ public function missingData(): array
             $missingData[] =  'Megjelölt kar';
         }
 
-        if (!$user->isResident() && !$user->isExtern()) {
+        if (!$user->resides() && !$user->isExtern()) {
```
kdmnk, line 292: same

app/Http/Controllers/Secretariat/SemesterEvaluationController.php
kdmnk, line 160: resides() is a scope, which I think returns a query builder (test it in tinker). Check if these checks work as expected. In the tests it also seems fine, just double check it, I would not call scopes this way. I would rather make it a simple function

app/Models/ApplicationForm.php
kdmnk, line 292: same

kdmnk (commented):
Can not we name this "LIMITED_RESIDENT" role internally? It is getting to be really confusing.
Alternatively, I was also thinking on making the roles time constrained. That way, we could have regular collegists, plus an additional resident role that is set to be valid for 1 year/indefinite.  We can also use that for guests, instead of the guest_until attribute and solve https://github.com/EotvosCollegium/mars/issues/291. (Also, the general collegist role is also valid only until the end of their studies by default which we can set this way).

@machiato32 what do you think, does putting effort into this would worth it?
(We have some time to work on this, the earliest this will be relevant is next summer)

kdmnk:
> Can not we name this "LIMITED_RESIDENT" role internally? It is getting to be really confusing. Alternatively, I was also thinking on making the roles time constrained. That way, we could have regular collegists, plus an additional resident role that is set to be valid for 1 year/indefinite. We can also use that for guests, instead of the guest_until attribute and solve #291. (Also, the general collegist role is also valid only until the end of their studies by default which we can set this way).
> 
> @machiato32 what do you think, does putting effort into this would worth it? (We have some time to work on this, the earliest this will be relevant is next summer)

I'm starting to like this idea. The role_user table would have a created_at and valid_until field, and we would use a global scope to filter on those. This way, we don't delete roles just invalidate them, making a log out of the table.

viktorcsimma:
> > Can not we name this "LIMITED_RESIDENT" role internally? It is getting to be really confusing. Alternatively, I was also thinking on making the roles time constrained. That way, we could have regular collegists, plus an additional resident role that is set to be valid for 1 year/indefinite. We can also use that for guests, instead of the guest_until attribute and solve #291. (Also, the general collegist role is also valid only until the end of their studies by default which we can set this way).
> > @machiato32 what do you think, does putting effort into this would worth it? (We have some time to work on this, the earliest this will be relevant is next summer)
> 
> I'm starting to like this idea. The role_user table would have a created_at and valid_until field, and we would use a global scope to filter on those. This way, we don't delete roles just invalidate them, making a log out of the table.

That sounds good:) Maybe we should open a new issue for that and freeze this until that gets solved.

viktorcsimma:
What should we do with this one? Should we merge it and then reverse it when working on that issue, or abandon it altogether?

kdmnk:
I'll close this for now.

PR #381, opened 2023-12-24
Adding simplification for Sail instructions

viktorcsimma:
I've found an easier way to add the correct privilege; I would add it here.


kdmnk (approved):
Why do we still have collegiumnostrum references here? 
Otherwise I don't really care here just merge :D
 and Merry Christmas 🤶

viktorcsimma:
Oh; I've missed that one.
And merry Christmas for you too;)

PR #384, opened 2023-12-24
Adding list of workshop functionaries to the home page

### LINKED ISSUE #363: Listing "Műhelytitkár"s in Urán for everyone
On the home page, in the contact section (or somewhere else in Urán where everyone can access it), display the list of "műhelytitkár"s like the members of the Students' Council. It's kinda useful to have all of them in a single list somewhere instead of having to look for them on the website, which is not always uptodate.

viktorcsimma:
I already have a functioning (yet ugly) version, which lists both the leaders and the administrators.


app/Models/Workshop.php
```diff
@@ -70,6 +71,34 @@ public function externs()
...
+    /**
+     * Filters functionaries to only include workshop leaders.
+     */
+    public function leaders()
+    {
+        return $this->functionaries->filter(function ($user, $key) {
+            return $user->hasRole(Role::WORKSHOP_LEADER);
+        });
+    }
+
```
kdmnk: The administrators and leaders should return relationships, not collection.
You can define the inverted version of roleWorkshops in the User model for each of them separately.

resources/views/home.blade.php
```diff
@@ -111,6 +111,48 @@
...
+                            <b>{{$workshop->name}}</b>
+                            <ul>
+                                <li>@lang('role.'.\App\Models\Role::WORKSHOP_LEADER):
+                                    <i>
+                                        {{
+                                            implode(
+                                                ', ',
+                                                array_map(
+                                                    function ($admin) {return $admin->name;},
+                                                    $workshop->leaders()->all()
```
kdmnk: $workshop->leaders->->pluck('name')->implode(', ');

kdmnk: Or it's better to calculate them in the backend and add it to the $contacts (see above), mapped to the workshops.

resources/views/home.blade.php
```diff
@@ -168,5 +210,18 @@ function standby(id) {
...
+var triangleSpan = document.getElementById("workshop-administrator-collapsible").firstElementChild;
+function toggleCollContent() {
+    if (collContent.style.display === "block") {
+        collContent.style.display = "none";
+        triangleSpan.textContent = "▶";
+    } else {
+        collContent.style.display = "block";
+        triangleSpan.textContent = "▼";
+    }
+}
```
kdmnk: BTW we have this: https://materializecss.com/collapsible.html
but not sure if it would look nice here. You can try some options before applying my other suggestions.

You can make this cleaner with jquery, this is generated with chatgpt:
```
  var collContent = $("#workshop-administrator-content");
  var triangleSpan = $("#workshop-administrator-collapsible").find(":first-child");

  function toggleCollContent() {
    if (collContent.is(":visible")) {
      collContent.hide();
      triangleSpan.text("▶");
    } else {
      collContent.show();
      triangleSpan.text("▼");
    }
  }

  $("#workshop-administrator-collapsible").on("click", toggleCollContent);
});
```

resources/views/home.blade.php
```diff
@@ -111,6 +111,48 @@
                     @endif
                 @endforeach
 
+                <!-- Workshop secretaries -->
+                <h5><a id="workshop-administrator-collapsible" href="javascript:void(0)" onclick="toggleCollContent()">
```
kdmnk: href="#"and you don't need the onclick with jquery (see below)

resources/sass/materialize.scss
```diff
@@ -68,6 +68,16 @@ span.badge.tag {
     float: none;
 }
 
+// for the list of workshop functionaries on the home page
+a#workshop-administrator-collapsible {
+    color: inherit; text-decoration: inherit;
+}
+.dropdown-triangle {
+    display: inline-block; font-size: 0.6em; transform: translateY(-0.4em);
+}
```
kdmnk: You can also add the icon content with CSS, so we don't have to specify it in HTML. 
You can also check if another class also exists on an element, it displays the other triangle.

You should be able to combine these to get a HTML structure like this. When you click on the div, the toggle functions toggles the "hide" class on the content while updating the icon. 
```
<div class="dropdown"> //display the corresponding icon, this has an event listener with the toggle function
<div class="dropdown-title">Title</div>
<div class="dropdown-content">Some content that may be hidden</div>
</div>
```

kdmnk: And I would add the icon after the title, it can mess up with the margins on the left side.

kdmnk: And if you do this, don't forget to add the .on('click', ...) listener in the main app layout page, next to the other initialisers.

kdmnk (changes requested):

kdmnk:
I've also tried the front end. Making it a link is a bit ugly, the black text with the icon is perfect enough, just make sure that the clickable cursor version is applied. And highlight the workshops more, or put the details of a workshop into one line or use a table.

resources/views/home.blade.php
viktorcsimma: I've tried Materialize collapsibles; but they looked odd; that's why I chose this one. But with your jquery solution, it works nicely.

viktorcsimma:
Is the current version OK? The workshop names are bold and roles have a left margin.

app/Http/Controllers/HomeController.php
```diff
@@ -218,7 +219,16 @@ private function getHomePageContacts(): array
...
                 Role::ETHICS_COMMISSIONER => User::ethicsCommissioners(),
             ]);
+
+            $workshops = Workshop::all();
+            $contacts['workshops'] = array_combine(
+                $workshops->pluck('name')->all(),
+                $workshops->map(fn ($workshop) => ['leaders' => $workshop->leaders->pluck('name')->implode(', '),
+                                                   'administrators' => $workshop->administrators->pluck('name')->implode(', ')])
+                          ->all()
+            );
```
kdmnk: Workshop::all()->flatMap(fn ($workshop) => [$workshop->name => ['leaders' => ..., 'administrators' => ...]])

resources/views/layouts/app.blade.php
```diff
@@ -103,6 +103,27 @@ function toggleColorMode() {
...
+                    title.addClass("closed");
+                }
+            }
+        }
+
+        $(".arrow-dropdown .arrow-dropdown-title").each(function () {
+            $(this).on('click', toggleCollContent($(this)));
+            // initialize it as closed:
+            $(this).addClass('closed');
+        });
```
kdmnk: I'm not sure if you need the each. Does not `$(".arrow-dropdown .arrow-dropdown-title").on('...')` work? But it's fine either way, I'm just curious.

kdmnk:
This looks really nice now, good job.

viktorcsimma:
Thanks:) Can we merge it now?

kdmnk (changes requested):
Ah sorry, I did not publish my review. Here it is, but nothing serious.

resources/views/layouts/app.blade.php
viktorcsimma: It does; in a strange way, but still. $(this) has to be substituted by the collection itself.

kdmnk (approved):

PR #386, opened 2023-12-26
Rework network module

### LINKED ISSUE #309: Internet page improvements
- [ ] Show the sum of other MAC addresses when displaying MAC address requests.
- [x] Show wifi username on internet access table

### LINKED ISSUE #370: Remove confusing characters from password generation
It has happened several times that users are having troubles connecting to the network. Usually it is only an issue of not typing the password correctly. Like confusing `y` with `z` (common English vs common Hungarian keyboard), or confusing `I` with `l` or `1`. We should probably look into the types of characters that can cause confusion and remove them from WiFi password generation.

I think it could drastically improve user experience without the users even noticing it.

kdmnk:
- Split internetcontroller into InternetController, AdminInternetController and MacAddressController.
- Deleted internet-user role
- Deleted auto-approved MACs for https://github.com/EotvosCollegium/mars/security/advisories/GHSA-x3j4-cf37-2fpr
- Tried to use REST structures
- Test for all the controller actions
- Improved tables on the pages
- Added wifi username and password to admin page
- #370 
- And some other changes that should have been in a separate PR, sorry: generated ide helpers, commented out triggers in migrations, and formatting
- New internet fault report form
- Improved email testing across test cases


app/Policies/MacAddressPolicy.php
```diff
@@ -16,57 +15,31 @@ public function before(User $user)
...
+     * Determine whether the user can create mac address instances.
      *
      * @param User $user
-     * @return mixed
+     * @return bool
      */
     public function create(User $user): bool
     {
-        return true;
+        return $user->can('handle', $user->internetAccess);
```
viktorcsimma: Does this mean that the user has an internet access? It is a bit confusing for me.

viktorcsimma: Maybe it would be better to sign it like, "if the user's internet access does exist and has not expired".

app/Models/Internet/InternetAccess.php
```diff
@@ -67,6 +120,34 @@ public function setWifiCredentials($username = null): string
...
+
+    /**
+     * Generate a "random" alphanumeric string, with the confusing characters filtered out.
+     * This is a modified version of Str::random() from Laravel.
+     *
+     * @return string
+     */
+    private static function generateWifiPassword(): string
+    {
+        $disabled_chars = ['/', '+', '=', 'o', '0', 'z', 'y', 'l', '1', 'i', 'I'];
```
viktorcsimma: Here, capital 'Z' and 'Y' are missing; I was able to reproduce them in generated passwords, too.

app/Models/Internet/InternetAccess.php
```diff
@@ -67,6 +120,34 @@ public function setWifiCredentials($username = null): string
...
+
+        $string = '';
+
+        while (($len = strlen($string)) < self::PASSWORD_LENGTH) {
+            $size = self::PASSWORD_LENGTH - $len;
+
+            $bytes = random_bytes($size);
+
+            $string .= substr(str_replace($disabled_chars, '', base64_encode($bytes)), 0, $size);
```
viktorcsimma, line 167: What does the `substr` call do here? If I understand right, this already is at most `$size` long.

app/Models/EpistolaNews.php
```diff
@@ -6,6 +6,56 @@
...
+ *
+ * @property int $id
+ * @property int|null $uploader_id
+ * @property string $title
+ * @property string|null $subtitle
+ * @property string|null $description
+ * @property \Illuminate\Support\Carbon|null $date
+ * @property \Illuminate\Support\Carbon|null $time
+ * @property \Illuminate\Support\Carbon|null $end_date
+ * @property string|null $picture_path
```
viktorcsimma, line 20: Just a question for myself; what are these?

viktorcsimma, line 20: Just a question for myself; what are these?

database/migrations/2019_10_19_153240_create_trigger.php
```diff
@@ -9,36 +9,36 @@ class CreateTrigger extends Migration
...
                ON users.id = print_accounts.user_id
             WHERE
                print_accounts.user_id IS NULL;');
-        DB::unprepared('
-        CREATE TRIGGER trigger_create_print_account_for_user
-            AFTER INSERT ON users 
-            FOR EACH ROW
-            INSERT INTO print_accounts(user_id) VALUES (NEW.id);
-        ');
+        //        DB::unprepared('
```
viktorcsimma, line 20: Maybe we should delete these comments altogether; we can still restore them from git if we need them.

viktorcsimma: Same at the other places.

database/migrations/2023_01_09_165653_drop_triggers.php
```diff
@@ -13,11 +13,11 @@
...
-        DB::unprepared('DROP TRIGGER trigger_insert_print_account_history_free_pages');
-        DB::unprepared('DROP TRIGGER trigger_update_print_account_history_free_pages');
+        //        DB::unprepared('DROP TRIGGER trigger_create_print_account_for_user');
+        //        DB::unprepared('DROP TRIGGER trigger_create_internet_access_for_user');
+        //        DB::unprepared('DROP TRIGGER trigger_print_account_history_balance');
+        //        DB::unprepared('DROP TRIGGER trigger_insert_print_account_history_free_pages');
+        //        DB::unprepared('DROP TRIGGER trigger_update_print_account_history_free_pages');
     }
 
     /**
```
viktorcsimma, line 23: Or is there any argument against simply deleting the entire migration file? If yes, we could add a different one reversing it (so readding the triggers).

resources/lang/hu/internet.php
```diff
@@ -19,13 +18,14 @@
...
+    'registered_devices_info' => 'A vezetékes internet használatához előbb regisztrálnod kell az eszközödet. Ezt biztonsági okokból manuálisan el kell fogadniuk a rendszergazdáknak, így szíves türelmedet kérjük. Igyekszünk minél hamarabb elfogadni a kérelmeket.',
     'rejected' => 'Elutasított',
     'report' => 'A probléma leírása',
     'report_fault' => 'Hiba bejelentése',
-    'report_fault_desc' => 'Ha bármi probléma merül fel az interneteléréseddel kapcsolatban, kérlek jelentsd be az alábbi űrlap kitöltésével. Próbáld meg a hibát minél pontosabban leírni és jelezd, mikor vagy a szobában, hogy a rendszergazdák fel tudjanak keresni.',
+    'report_fault_desc' => 'Ha bármi probléma merül fel az interneteléréseddel kapcsolatban, kérlek jelentsd be az alábbi űrlap kitöltésével. Próbáld meg a hibát minél pontosabban leírni (mióta, hogyan jelentkezik a probléma, van-e hibaüzenet, próbáltad-e más eszközön, stb.) és jelezd, mikor vagy a szobában, hogy a rendszergazdák fel tudjanak keresni.',
```
viktorcsimma: We could also add something like "what you have already tried to solve the issue". This would save us from many unnecessary rounds.

resources/lang/en/internet.php
```diff
@@ -19,9 +18,10 @@
     'internet_access' => 'Internet access',
     'laptop' => 'Laptop',
     'mac_address' => 'MAC address',
+    'mac_comment' => 'eg. Asus Laptop, Ethernet converter',
     'no_internet' => 'You are not allowed to access the internet at the moment.',
     'nothing_to_show' => 'Nothing to show',
-    'registered_devices_info' => 'You must register your device to use wired connection. You can approve a certain number of devices automatically. If you need more, then the administrators must approve it after your registration; you can provide a reason in the comment field.',
+    'registered_devices_info' => 'To use wired internet, you must first register your device. For security reasons, system administrators have to accept this manually, so we kindly asks for your patience. We try to accept requests as soon as possible.',
```
viktorcsimma: Just a grammatical error: `kindly ask`.

resources/lang/hu/internet.php
viktorcsimma: Something else: should we also extend this text in the English version, too?

resources/views/network/admin/mac_addresses.blade.php
```diff
@@ -0,0 +1,150 @@
+<span
+    class="card-title">Regisztrált vezetékes eszközök kezelése
+    @notification(\App\Models\Internet\MacAddress::class)
+</span>
+<blockquote class="error">
+    A MAC címek jóváhagyása előtt ellenőrizd, hány címe van már a felhasználónak, és hogy nem routerről van-e szó
+    (<a href="https://mac-address.alldatafeeds.com/mac-address-lookup" target="_blank">ellenőrizd pl. itt</a>).
+    Módosítsd a megjegyzést, ha szükséges.
```
viktorcsimma: We could include some kind of warning about what @horcsinbalint has said about `ff:ff:ff:ff:ff` and friends.

app/Models/Role.php
```diff
@@ -293,7 +306,7 @@ public function color(): string
             self::LOCALE_ADMIN => 'amber',
```
viktorcsimma, line 305: Here, `self::INTERNET_USER` is still there; and it complains about that when I try to access the users index.

viktorcsimma (changes requested):
It seems to be nice:) One important thing besides my comments: I get a 403 when I try to access the routers' index. Could you investigate that?

Btw, thanks for trusting me with the request:) sorry if I ask something that I should know.

app/Policies/MacAddressPolicy.php
kdmnk: This only redirects it to the InternetAccessPolicy's handle which only checks currently that it can not create a Mac for other users. Which is true in every case in this call :D I'll just delete it. (The internet access should exist by default, but if not it also returns 403 for null. The expired check is not that necessary, just creates more tension.)

app/Models/Internet/InternetAccess.php
kdmnk: idk, as I said in the comments, I just modified laravel's algorithm, which I trust :D

app/Models/EpistolaNews.php
kdmnk: These should help your IDE find the Laravel-generated properties and their types. These are generated with PHP artisan ide-helper (an external tool)

database/migrations/2019_10_19_153240_create_trigger.php
kdmnk: Our git history is not the nicest the search in... They won't harm us here ;) And we can squash the migrations into one file now containing a final db structure, I just could not make it work yet.

app/Http/Controllers/Network/AdminInternetController.php
```diff
@@ -0,0 +1,111 @@
...
+     * @return LengthAwarePaginator
+     * @throws AuthorizationException
+     */
+    public function indexInternetAccesses(): LengthAwarePaginator
+    {
+        $this->authorize('handleAny', InternetAccess::class);
+
+        return TabulatorPaginator::from(
+            InternetAccess::query()
+                ->join('users as user', 'user.id', '=', 'user_id')
```
sszajbely, line 45: Is this `join` needed when using `with` below?

app/Http/Controllers/Network/AdminInternetController.php
```diff
@@ -0,0 +1,111 @@
...
+        ]);
+
+        if ($request->get('has_internet_until') != null) {
+            $newDate = $request->get('has_internet_until');
+        } else {
+            $newDate = InternetAccess::getInternetDeadline();
+        }
+        $internetAccess->update(['has_internet_until' => $newDate]);
+        return Carbon::parse($newDate);
+    }
```
sszajbely: Maybe this instead?
```suggestion
        if ($request->get('has_internet_until') != null) {
            $newDate = Carbon::parse($request->get('has_internet_until'));
        } else {
            $newDate = InternetAccess::getInternetDeadline();
        }
        $internetAccess->update(['has_internet_until' => $newDate]);
        return $newDate;
    }
```

app/Http/Controllers/Network/InternetController.php
```diff
@@ -4,250 +4,68 @@
...
+        $request->validate([
             'report' => 'required|string',
             'user_os' => 'required|string',
         ]);
-        $validator->validate();
 
-        foreach (User::withRole(Role::SYS_ADMIN)->get() as $admin) {
-            Mail::to($admin)->queue(new InternetFault($admin->name, user()->name, $request->report, $request->user_os));
+        foreach (User::admins() as $admin) {
+            Mail::to($admin)->queue(new InternetFault($admin->name, user()->name, $request->get('report'), $request->get('user_os')));
```
sszajbely: Question for all of these kind of emails. Do we really need the emails stating the name of the admin? Couldn't it just say 'Dear Admin, ...'? If that was okay we could use `Mail::to(User::admins())->...` instead.

app/Http/Controllers/Network/MacAddressController.php
```diff
@@ -0,0 +1,99 @@
...
+     * @return LengthAwarePaginator
+     * @throws AuthorizationException
+     */
+    public function index(Request $request): LengthAwarePaginator
+    {
+        $this->authorize('handleAny', InternetAccess::class);
+
+        return TabulatorPaginator::from(
+            MacAddress::query()
+                ->join('users as user', 'user.id', '=', 'user_id')
```
sszajbely, line 35: Same as above

app/Http/Controllers/Network/MacAddressController.php
```diff
@@ -0,0 +1,99 @@
...
+     * @throws ValidationException
+     */
+    public function store(Request $request): RedirectResponse
+    {
+        $this->authorize('create', MacAddress::class);
+
+        Validator::validate($request->all(), [
+            'comment' => 'required|max:255',
+            'mac_address' => ['required', 'regex:/((([a-fA-F0-9]{2}[-:]){5}([a-fA-F0-9]{2}))|(([a-fA-F0-9]{2}:){5}([a-fA-F0-9]{2})))/i'],
+        ]);
```
sszajbely:
```suggestion
        $request->validate([
            'comment' => 'required|max:255',
            'mac_address' => ['required', 'regex:/((([a-fA-F0-9]{2}[-:]){5}([a-fA-F0-9]{2}))|(([a-fA-F0-9]{2}:){5}([a-fA-F0-9]{2})))/i'],
        ]);
```

app/Http/Controllers/Network/MacAddressController.php
```diff
@@ -0,0 +1,99 @@
...
+
+        MacAddress::create([
+            'user_id' => user()->id,
+            'comment' => $request->input('comment'),
+            'mac_address' => $request->input('mac_address')
+        ]);
+
+        foreach (User::admins() as $admin) {
+            Mail::to($admin)->send(new MacNeedsApproval($admin->name, user()->name));
+        }
```
sszajbely: Same as above

app/Policies/MacAddressPolicy.php
```diff
@@ -16,57 +15,31 @@ public function before(User $user)
...
     /**
-     * Determine whether the user can update the mac address.
+     * Determine whether the user can approve a mac address.
      *
      * @param User $user
-     * @param MacAddress $macAddress
      * @return mixed
      */
-    public function update(User $user, MacAddress $macAddress): bool
+    public function updateState(User $user): bool
```
sszajbely: Why did you rename this?

app/Http/Controllers/Network/MacAddressController.php
```diff
@@ -0,0 +1,99 @@
...
+        $this->authorize('updateState', $macAddress);
+
+        $request->validate([
+            'comment' => 'nullable|string|max:255',
+            'state' => 'nullable|in:' . implode(',', MacAddress::STATES)
+        ]);
+
+        $macAddress->update($request->only(['state', 'comment']));
+
+        return $macAddress->refresh();
```
sszajbely: I don't think the refresh is needed (maybe I'm wrong)

app/Http/Controllers/Network/MacAddressController.php
```diff
@@ -0,0 +1,99 @@
...
+    public function update(Request $request, MacAddress $macAddress): MacAddress
+    {
+        $this->authorize('updateState', $macAddress);
+
+        $request->validate([
+            'comment' => 'nullable|string|max:255',
+            'state' => 'nullable|in:' . implode(',', MacAddress::STATES)
+        ]);
+
+        $macAddress->update($request->only(['state', 'comment']));
```
sszajbely:
```suggestion
        $data = $request->validate([
            'comment' => 'nullable|string|max:255',
            'state' => 'nullable|in:' . implode(',', MacAddress::STATES)
        ]);

        $macAddress->update($data);
```
It's a bit less code, but maybe less readable.

app/Models/Internet/InternetAccess.php
```diff
@@ -2,45 +2,98 @@
...
+ * @method static Builder|InternetAccess whereAutoApprovedMacSlots($value)
+ * @method static Builder|InternetAccess whereCreatedAt($value)
+ * @method static Builder|InternetAccess whereHasInternetUntil($value)
+ * @method static Builder|InternetAccess whereUpdatedAt($value)
+ * @method static Builder|InternetAccess whereUserId($value)
+ * @method static Builder|InternetAccess whereWifiPassword($value)
+ * @method static Builder|InternetAccess whereWifiUsername($value)
+ * @property-read Collection|MacAddress[] $macAddresses
+ * @property-read int|null $mac_addresses_count
+ * @mixin Eloquent
```
sszajbely, line 46: These IDE helpers would have been nice in a separate PR :D

app/Models/Internet/InternetAccess.php
```diff
@@ -2,45 +2,98 @@
...
  */
 class InternetAccess extends Model
 {
+    protected $table = 'internet_accesses';
```
sszajbely: Is this our convention to state the table name in each model? Because this is not needed otherwise.

app/Models/Internet/MacAddress.php
```diff
@@ -33,15 +55,49 @@ class MacAddress extends Model
         'user_id', 'mac_address', 'comment', 'state', 'ip',
     ];
 
+    // Default attributes
     protected $attributes = [
-        'comment' => '',
         'state' => self::REQUESTED,
     ];
```
sszajbely: What does this do exactly? Also it would be nice to use enums instead of const strings.

app/Models/Internet/MacAddress.php
```diff
@@ -33,15 +55,49 @@ class MacAddress extends Model
...
     {
-        return $this->belongsTo(User::class);
+        return $this->hasOneThrough(
+            User::class,
+            InternetAccess::class,
+            'user_id', // Foreign key on InternetAccess table...
+            'id', // Foreign key on Users table...
+            'user_id', // Local key on MacAddresses table...
+            'user_id' // Local key on InternetAccess table...
+        );
```
sszajbely, line 96: Why? Do you want to treat the `MacAddress` as if it belonged to the `InternetAccess` even though it is not reflected in the database?

app/Models/Internet/MacAddress.php
```diff
@@ -57,7 +113,11 @@ public function macAddress(): Attribute
         );
     }
 
-    public static function notifications()
+    /**
+     * The number of mac addresses waiting for approval.
+     * @return int
+     */
+    public static function notifications(): int
```
sszajbely: Could the name reflect the usage? Even something like `notificationCount`.

app/Models/Internet/Router.php
```diff
@@ -33,37 +70,30 @@ class Router extends Model
...
+     * Email the sysadmins and the residents if the router has failed for the second time.
+     * @return void
+     */
+    public function sendWarning(): void
     {
         if ($this->failed_for == self::WARNING_THRESHOLD) {
             foreach (User::admins() as $admin) {
-                Mail::to($admin)->queue(new \App\Mail\RouterWarning($admin, $this));
+                Mail::to($admin)->queue(new RouterWarning($admin, $this));
             }
```
sszajbely, line 106: Same as above

app/Models/Internet/Router.php
```diff
@@ -33,37 +70,30 @@ class Router extends Model
...
         }
     }
 
-    public static function notifications()
+    /**
+     * Return the number of routers that have failed.
+     * @return int
+     */
+    public static function notifications(): int
```
sszajbely: Same as above

database/migrations/2023_12_25_190920_delete_internet_access_role.php
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        if (Role::where('name', 'internet-user')->first() !== null) {
+            DB::table('roles')->delete(Role::where('name', 'internet-user')->first()->id);
+        }
```
sszajbely, line 18:
```suggestion
        if (Role::firstWhere('name', 'internet-user') !== null) {
            DB::table('roles')->delete(Role::firstWhere('name', 'internet-user')->id);
        }
```

Or maybe even `DB::table('roles')->delete(Role::firstWhere('name', 'internet-user')?->id);` without the first if. I don't know what delete with null as id does.

sszajbely (changes requested):
I didn't really look at the frontend, as I'm not really familiar with the blades. Other than my comments LGTM.

app/Http/Controllers/Network/AdminInternetController.php
kdmnk: I've tried but it seems like we need both for some reason. I haven't checked the TabulatorPaginator how we could optimise.

app/Http/Controllers/Network/MacAddressController.php
kdmnk: If the request has mac_address or user id, this would also update that, which is a security risk.

app/Models/Internet/InternetAccess.php
kdmnk: well yes, sorry

app/Models/Internet/InternetAccess.php
kdmnk: Sometimes we need it, but true, not here.

app/Models/Internet/MacAddress.php
kdmnk: Default values are defined here instead of DB level. We don't have a convention here. I created a new issue for enums as we need to change it everywhere.

app/Http/Controllers/Network/MacAddressController.php
sszajbely: Then the validation should reflect what you expect. $data will only contain data which was validated.

app/Models/Internet/MacAddress.php
kdmnk: Yes. I know its a bit strange here but I don't think this would cause problems

app/Models/Internet/MacAddress.php
sszajbely: But you can still do it here. Less to refactor later :D

app/Models/Internet/MacAddress.php
sszajbely: It's a bit strange and probably heavier on the queries, but I don't think it is a problem. It mainly was just a question.

app/Http/Controllers/Network/MacAddressController.php
```diff
@@ -0,0 +1,99 @@
...
+        ]);
+
+        MacAddress::create([
+            'user_id' => user()->id,
+            'comment' => $request->input('comment'),
+            'mac_address' => $request->input('mac_address')
+        ]);
+
+        foreach (User::admins() as $admin) {
+            Mail::to($admin)->send(new MacNeedsApproval($admin->name, user()->name));
```
horcsinbalint: Is it possible to change in a way that new MAC addresses from admins would be automatically approved? I think we might need to sometimes test out random devices and spamming other administrators' e-mail addresses would be painful.

sszajbely (approved):

PR #393, opened 2024-01-03
Merging recent changes into `deployment`

viktorcsimma:
I think there has been quite a few changes now to deploy them to the server. @kdmnk what do you think?

@horcsinbalint where should I pull on the server when this is merged?


viktorcsimma:
Huh; there are lots of warnings about documentation comments and other things... What should we do about them? The problem is that I've only written a small portion of these functions; so I don't know enough about them to write a documentation comment.

kdmnk:
I'll take a look later. First, the deployment branch is ahead of development, which should not be the case. I also added protection rules for deployment

kdmnk:
Or you make a PR from deployment to development first, and we update and start using again the abandoned `release` branch

horcsinbalint:
I agree with @kdmnk , we should merge the `deployment` branch into `development`, and fix the problems.
The general (semi-documented) steps are to update mars -- most steps are to make sure that data is not lost:

- Make a backup of all the data
- Set up password for mars in the reverse proxy:
  - Update reverse proxy's config on the other repo
  - Pull it on the server
  - Fix the permissions
  - Restart the reverse proxy
- Update the docker images
- Update the containers
- Run migrations
- Check if the website is running nicely
- Remove the password for mars in the reverse proxy:
  - Update reverse proxy's config on the other repo
  - Pull it on the server
  - Fix the permissions
  - Restart the reverse proxy

The method without safeguards (data-loss is possible, and I don't really like it):

- Update the docker images
- Update the containers
- Run migrations

Yeah, this is probably enough reason, why we need a staging environment and proper CI/CD. :D (With those we might not even need to interact with the server via SSH to make updates on mars.) I'm planning on implementing that, since having such a complicated deployment strategy will cause some disasters even short-term.

Can changes with the migrations about the triggers cause inconsistency between the server and development environments? (I have seen some changes in old migrations with statements `DB::unprepared(...)`.)

viktorcsimma:
> Or you make a PR from deployment to development first, and we update and start using again the abandoned `release` branch

That's a good idea; let's see what happens then;)

horcsinbalint:
I think the workshop functionaries list should be a dropdown list (closed by default). I guess the code tries to achieve it, but it does not work on my machine (Linux Mint 21.2 Cinnamon, Firefox 121.10, Chrome 120.0.6099.199). A quick fix could be to add it in details - summary tags. (https://www.w3schools.com/tags/tag_summary.asp)

horcsinbalint:
> > Or you make a PR from deployment to development first, and we update and start using again the abandoned `release` branch
> 
> That's a good idea; let's see what happens then;)

I would like to keep temporarily the `deployment` branch, because its name is hardcoded in the other repo. We have to do quite some refactoring on the operational side, and I would rather do the namechange when I create the infrastructure for the `staging` branch.

kdmnk:
> I think the workshop functionaries list should be a dropdown list (closed by default). I guess the code tries to achieve it, but it does not work on my machine (Linux Mint 21.2 Cinnamon, Firefox 121.10, Chrome 120.0.6099.199). A quick fix could be to add it in details - summary tags. (https://www.w3schools.com/tags/tag_summary.asp)

@horcsinbalint you need to run `npm run prod` if we update any CSS.

PR #394, opened 2024-01-03
Rework print module

### LINKED ISSUE #51: Free pages double sided printing
If you have 12 pages you should be able to print a 24 page long document if you choose to use double sided printing. Right it does not work.

### LINKED ISSUE #344: Set free printing true by default
Fwd: "Meg lehetne oldani azt esetleg, hogy akinek van ingyen nyomtatása, annak ne kelljen beállítgatni az ingyen oldalak nyomtatását, hanem automatikusan abból vonja le neki az oldalakat?"

### LINKED ISSUE #366: Support external CUPS server
It is sometimes better to have a CUPS server that is located on a different server. (Like when the CUPS server is running nearby the printer on a different floor.) There is currently no supported method of setting the IP address and port of the CUPS server for commands `lp` and `lpstat` in mars. It could be beneficial to add an additional field in .env which allows arbitrary command line arguments to be inserted in front of the `-d` and `-W` directive respectively.

### LINKED ISSUE #374: Allows either one- or two-sided printing of 2-page document with 1 free page
(but correctly subtracts free pages when you have enough)

sszajbely:
- Split up `PrintController` into `PrinterController`, `PrintJobController`, `PrintAccountController`, `PrintAccountHistoryController`, `FreePagesController`.
- Tried to use REST standards wherever it was possible.
- Removed `Printer` utility class, and printing related commands from the `Commands` utility.
- Removed printer role
- Fixed some bugs regarding paginated tables.
- Added `printers` table. This is mostly done for better REST compatibility and cleaner code, but still is useful if (when) multiple printers are installed at some point.
- Documents are now deleted after they are sent to the printer.

Closes #51, #374, #366 and #344


sszajbely:
I still have to write the tests, but the main refactoring is done.

.vscode/settings.json
horcsinbalint, line 1: I think this file could potentially be removed from git and added to .gitignore

sszajbely: I don't even know why that's there. Good catch

app/Console/Commands/CreateDefaultPrinter.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Models\Printer;
+use Illuminate\Console\Command;
+
+class CreateDefaultPrinter extends Command
```
kdmnk: Why do you need this? If needed, then let's make this general so we can provide the routers attributes in command arguments.

app/Http/Controllers/Dormitory/Printing/FreePagesController.php
```diff
@@ -0,0 +1,83 @@
...
+use Illuminate\Pagination\LengthAwarePaginator;
+
+class FreePagesController extends Controller
+{
+    /**
+     * Returns a paginated list of `FreePages`.
+     * @param null|string $filter Decides wether all `FreePages` or just the user's `FreePages` should be listed.
+     * @return LengthAwarePaginator
+     */
+    public function indexFreePages(?string $filter = null)
```
kdmnk: Call this only 'index' and do not handle the not admin part here. Modify the table in the view to use data passed with the normal print index page's data. That's what I did with the mac addresses, you can check. I know it's too simple but we should follow that here then.

app/Http/Controllers/Dormitory/Printing/PrintAccountController.php
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Updates balance of a `PrintAccount`.
+     * This can be done in two ways: either by topping up the account by giving money to an admin
+     * or by transfering money from one account to the other.
+     */
+    public function update(Request $request)
+    {
+        $request->validate([
+            'amount' => 'required|integer',
+            'user' => 'required|exists:users,id', // Normally this would be a path parameter for the PrintAccount, but we can't do that because of the limitations of blade templates
```
kdmnk, line 32: Why? For the select input, you should be able to pass the printaccount id's as keys.

app/Http/Controllers/Dormitory/Printing/PrintAccountController.php
```diff
@@ -0,0 +1,128 @@
...
+            'other_user' => 'nullable|exists:users,id',
+        ]);
+
+        $printAccount = User::find($request->get('user'))->printAccount;
+
+        // If user can not even transfer balance, we can stop here
+        if (user()->cannot('transferBalance', $printAccount)) {
```
kdmnk: Why not use $this->authorize? (everywhere)

app/Http/Controllers/Dormitory/Printing/PrintAccountController.php
```diff
@@ -0,0 +1,128 @@
...
+            }
+
+            // Cannot transfer if there is not enough balance to be transfered
+            if ($printAccount->balance < $amount) {
+                return $this->returnNoBalance();
+            }
+
+            $printAccount->update([
+                'balance' => $printAccount->balance - $amount,
+                'last_modified_by' => user()->id,
```
kdmnk: Does it make sense to store the last modified by? We should send an email every time with the actors stated. That should be enough, shouldn't it? Then you could just use increment/decrement. Or if we keep it, move that setter to an observer.

app/Http/Controllers/Dormitory/Printing/PrintAccountController.php
```diff
@@ -0,0 +1,128 @@
...
+                'balance' => $otherAccount->balance + $amount,
+                'last_modified_by' => user()->id,
+            ]);
+
+            Mail::to($printAccount->user)->queue(new ChangedPrintBalance($printAccount->user, $request->get('amount'), user()->name));
+
+            return redirect()->back()->with('message', __('general.successful_transaction'));
+        }
+        // This is a modification of the current account
+        else {
```
kdmnk: Nah, I would rather separate the endpoints here.

app/Http/Controllers/Dormitory/Printing/PrintAccountHistoryController.php
```diff
@@ -0,0 +1,34 @@
...
+class PrintAccountHistoryController extends Controller
+{
+    /**
+     * Returns a paginated list of `PrintAccountHistory` entries.
+     * @return LengthAwarePaginator
+     * @throws AuthorizationException
+     * @throws BindingResolutionException
+     * @throws InvalidArgumentException
+     */
+    public function indexPrintAccountHistory()
```
kdmnk: Call this index

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+use Storage;
+
+class PrintJobController extends Controller
+{
+    /**
+     * Returns a paginated list of `PrintJob`s.
+     * @param null|string $filter Decides wether all `PrintJob`s or just the user's `PrintJob`s should be listed.
+     * @return LengthAwarePaginator
+     */
+    public function indexPrintJobs(?string $filter = null)
```
kdmnk: Call this index

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+
+    /**
+     * Returns a paginated list of `PrintJob`s.
+     * @param Builder $printJobs
+     * @param array $columns
+     * @return LengthAwarePaginator
+     * @throws BindingResolutionException
+     * @throws InvalidArgumentException
+     */
+    private function paginatorFrom(Builder $printJobs, array $columns)
```
kdmnk, line 173: If we always use this logic, why don't you modify the api for the tabulatorpaginator?

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+class PrintJobController extends Controller
+{
+    /**
+     * Returns a paginated list of `PrintJob`s.
+     * @param null|string $filter Decides wether all `PrintJob`s or just the user's `PrintJob`s should be listed.
+     * @return LengthAwarePaginator
+     */
+    public function indexPrintJobs(?string $filter = null)
+    {
+        if ($filter === "all") {
```
kdmnk: Split this into 2 functions or create a new AdminPrintJobController

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    /**
+     * Prints a document, then stores the corresponding `PrintJob`.
+     * @param Request $request
+     * @return RedirectResponse
+     */
+    public function store(Request $request)
+    {
+        xdebug_break();
```
kdmnk: Delete

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+        Log::info($request->all());
```
kdmnk: Delete

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+        $request->validate([
+            'file' => 'required|file',
+            'copies' => 'required|integer|min:1',
+            'two_sided' => 'in:on,off',
+            'printer_id' => 'exists:printers,id',
+            'use_free_pages' => 'in:on,off',
+        ]);
+        Log::info("Printjob cost 2");
```
kdmnk: Delete...

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+
+        $useFreePages = $request->boolean('use_free_pages');
+        $copyNumber = $request->input('copies');
+        $twoSided = $request->boolean('two_sided');
+        $file = $request->file('file');
+
+        /** @var Printer */
+        $printer = $request->has('printer_id') ? Printer::find($request->input("printer_id")) : Printer::firstWhere('name', config('print.printer_name'));
```
kdmnk: Why are there 2 different ways to provide the printer? Can't you move the ID to the url?

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+        $request->validate([
+            'file' => 'required|file',
+            'copies' => 'required|integer|min:1',
+            'two_sided' => 'in:on,off',
+            'printer_id' => 'exists:printers,id',
+            'use_free_pages' => 'in:on,off',
+        ]);
+        Log::info("Printjob cost 2");
+
+        $useFreePages = $request->boolean('use_free_pages');
```
kdmnk, line 90: Only keep this variables if you use them multiple times (I can't see now from mobile :D)

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+        $printAccount = user()->printAccount;
+
+        if (!(($useFreePages && $printAccount->hasEnoughFreePages($pageNumber, $copyNumber, $twoSided)) ||
+            (!$useFreePages && $printAccount->hasEnoughBalance($pageNumber, $copyNumber, $twoSided)))
+        ) {
+            return back()->with('error', __('print.no_balance'));
+        }
+
+        $jobId = null;
+        try {
```
kdmnk, line 116: Can't you move the try catch inside the function? 

$jobId = $printer->print...
Storage::delete($path)
if (jobId == null) return back()...

app/Http/Controllers/Dormitory/Printing/PrinterController.php
```diff
@@ -0,0 +1,61 @@
...
+class PrinterController extends Controller
+{
+    /**
+     * Returns the print page for either the admin or the normal page.
+     * @param null|string $page Determines wether we are on the admin or the normal printing page.
+     * @return View
+     */
+    public function index(?string $page = null)
+    {
+        if ($page === "admin") {
```
kdmnk: Create AdminPrintController or split it into two functions

app/Http/Controllers/Dormitory/Printing/PrintAccountController.php
```diff
@@ -0,0 +1,128 @@
...
+
+            // Cannot transfer from other user's account (even if you are admin)
+            if ($printAccount->user_id !== user()->id) {
+                abort(403);
+            }
+
+            $amount = $request->get('amount');
+
+            // This would be effectively stealing printing money from the other account
+            if ($amount < 0) {
```
kdmnk: move this to validation

app/Http/Controllers/Dormitory/Printing/PrintAccountController.php
```diff
@@ -0,0 +1,128 @@
...
+        if (user()->cannot('transferBalance', $printAccount)) {
+            abort(403);
+        }
+
+        $otherAccount = $request->other_user ? User::find($request->get('other_user'))->printAccount : null;
+
+        // This is a transfer between accounts
+        if ($otherAccount !== null) {
+            // Cannot transfer to yourself
+            if ($otherAccount->user_id === $printAccount->user_id) {
```
kdmnk: why not? keep this simple

app/Http/Controllers/Dormitory/Printing/PrintAccountController.php
```diff
@@ -0,0 +1,128 @@
...
+                abort(400);
+            }
+
+            // Cannot transfer from other user's account (even if you are admin)
+            if ($printAccount->user_id !== user()->id) {
```
kdmnk: Don't need this if you separate the endpoint (see below)

app/Http/Controllers/Dormitory/Printing/PrintAccountController.php
```diff
@@ -0,0 +1,128 @@
...
+        else {
+            // Only admins can modify accounts
+            if (user()->cannot('modify', $printAccount)) {
+                abort(403);
+            }
+
+            $amount = $request->get('amount');
+
+            if ($amount < 0 && $printAccount->balance < $amount) {
+                $this->returnNoBalance();
```
kdmnk: are you sure this will return from the function here?

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+        $pageNumber = Printer::getDocumentPageNumber($path);
+
+        /** @var PrintAccount */
+        $printAccount = user()->printAccount;
+
+        if (!(($useFreePages && $printAccount->hasEnoughFreePages($pageNumber, $copyNumber, $twoSided)) ||
+            (!$useFreePages && $printAccount->hasEnoughBalance($pageNumber, $copyNumber, $twoSided)))
+        ) {
+            return back()->with('error', __('print.no_balance'));
+        }
```
kdmnk: Can you move both cases into the hasEnoughBalance? (and pass the useFreePages var). Or just make the condition clearer, it's hard to see what is happening with all the negations. Or move this to custom validation rules.

kdmnk: Also, you are calling getFreePagesNeeded/getBalanceNeeded later, which seems like a similar logic.

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+            return back()->with('error', __('print.error_printing'));
+        } finally {
+            Storage::delete($path);
+        }
+
+        $cost = $useFreePages ?
+            PrintAccount::getFreePagesNeeeded($pageNumber, $copyNumber, $twoSided) :
+            PrintAccount::getBalanceNeeded($pageNumber, $copyNumber, $twoSided);
+
+        Log::info("Printjob cost: $cost");
```
kdmnk: Make one log message that has all the relevant info, if you need. A printjob cost entry on its own does not say much.

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+                }
+            }
+            // Set value in the session so that free page checkbox stays checked
+            session()->put('use_free_pages', true);
+        } else {
+            $printAccount->balance -= $cost;
+
+            // Remove value regarding the free page checkbox from the session
+            session()->remove('use_free_pages');
+        }
```
kdmnk: This logic should be elsewhere. Create unit tests for this, if you don't have yet.

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+
+        return back()->with('message', __('print.success'));
+    }
+
+    /**
+     * Cancels a `PrintJob`
+     * @param PrintJob $job
+     * @return RedirectResponse
+     */
+    public function update(PrintJob $job)
```
kdmnk: Then name it cancel. Or if you want to stick with restful names, in terms of print jobs, this should be 'destroy' with a soft delete, which makes sense to me.

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+
+        if (!(($useFreePages && $printAccount->hasEnoughFreePages($pageNumber, $copyNumber, $twoSided)) ||
+            (!$useFreePages && $printAccount->hasEnoughBalance($pageNumber, $copyNumber, $twoSided)))
+        ) {
+            return back()->with('error', __('print.no_balance'));
+        }
+
+        $jobId = null;
+        try {
+            $jobId = $printer->print($twoSided, $copyNumber, $path);
```
kdmnk: Let's call this 'startPrintJob', if this is supposed to be that,

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+     * Cancels a `PrintJob`
+     * @param PrintJob $job
+     * @return RedirectResponse
+     */
+    public function update(PrintJob $job)
+    {
+        $this->authorize('update', $job);
+
+        if ($job->state === PrintJobStatus::QUEUED) {
+            $result = ($job->printer ?? Printer::firstWhere('name', config('print.printer_name')))->cancelPrintJob($job);
```
kdmnk: same, why the printer is not required for a job?

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,231 @@
...
+
+                    if ($job->used_free_pages) {
+                        $pages = $printAccount->available_free_pages->first();
+                        $pages->update([
+                            'last_modified_by' => user()->id,
+                            'amount' => $pages->amount + $job->cost,
+                        ]);
+                    } else {
+                        $printAccount->balance += $job->cost;
+                    }
```
kdmnk: This should happen automatically, through observers. If you use soft delete, you can also use the restore event here. And make a function for the printAcount to add/subtract free pages from/to the right free pages model. You should not worry about the free pages representation here.

app/Models/PrintAccount.php
```diff
@@ -54,31 +57,101 @@ class PrintAccount extends Model
...
-        return $this->hasMany('App\Models\FreePages', 'user_id', 'user_id');
+        return $this->hasMany(FreePages::class, 'user_id', 'user_id');
     }
 
-    public function hasEnoughMoney($balance)
+    /**
+     * The free pages which are currently available. Sorts the free pages by their deadline.
+     * @return Collection
+     */
+    public function getAvailableFreePagesAttribute()
```
kdmnk: Use Laravel's Attribute class. You can see an example in MacAddress. Or, rather make this a more constrained hasMany relation getter without the '->get()' at the end.

app/Models/PrintAccount.php
```diff
@@ -54,31 +57,101 @@ class PrintAccount extends Model
...
 
-        return $one_copy_cost * $number_of_copies;
+    /**
+     * Returns the number of free pages needed to print with given configuration.
+     * @param int $pages
+     * @param mixed $copies
+     * @param mixed $twoSided
+     * @return int|float
+     */
+    public static function getFreePagesNeeeded(int $pages, $copies, $twoSided)
```
kdmnk: call this 'getPagesNeeded'

app/Models/PrintJob.php
```diff
@@ -37,46 +41,79 @@ class PrintJob extends Model
...
+            'user_id', // Local key on print_jobs
+            'id', // Local key on users
+        );
     }
 
-    public static function addCurrencyTag(): \Closure
+    /**
+     * Attribute for the translated cost.
+     */
+    public function getTranslatedCostAttribute()
```
kdmnk: Use Attributes. (See MacAddress)

app/Models/PrintJob.php
```diff
@@ -37,46 +41,79 @@ class PrintJob extends Model
...
-            $data->cost = "{$data->cost} HUF";
+        return $this->used_free_pages ? "$this->cost ingyenes oldal" : "$this->cost HUF";
+    }
 
-            return $data;
-        };
+    /**
+     * Attribute for the translated state.
+     */
+    public function getTranslatedStateAttribute()
```
kdmnk: Same

app/Models/PrintJob.php
```diff
@@ -37,46 +41,79 @@ class PrintJob extends Model
...
+        return __("print." . strtoupper($this->state->value));
+    }
+
+    /**
+     * Gets the printjob-status with every printer, updates the status of the completed printjobs.
+     */
+    public static function checkAndUpdateStatuses()
+    {
+        foreach(PrintJob::query()->where('state', PrintJobStatus::QUEUED)->whereNotNull('printer_id')->pluck('printer_id')->unique() as $printer_id) {
+            $printJobs = Printer::find($printer_id)->getCompletedPrintJobs();
```
kdmnk: Can't you get failed jobs or warnings also?

app/Models/Printer.php
```diff
@@ -0,0 +1,154 @@
...
+            $result = ['output' => '', 'exit_code' => 0];
+        } else {
+            $process->run();
+            $result = ['output' => $process->getOutput(), 'exit_code' => $process->getExitCode()];
+        }
+        Log::info([$process->getCommandLine(), $result]);
+        if ($result['exit_code'] == 0) {
+            return PrinterCancelResult::Success;
+        }
+        if (strpos($result['output'], "already canceled") !== false) {
```
kdmnk: isn't this a typo?

app/Models/Printer.php
```diff
@@ -0,0 +1,154 @@
...
+            $result = $process->getOutput();
+            if (!preg_match("/^request id is ([^\s]*) \\([0-9]* file\\(s\\)\\)$/", $result, $matches)) {
+                Log::error("Printing error at line: " . __FILE__ . ":" . __LINE__ . " (in function " . __FUNCTION__ . "). result:"
+                    . print_r($result, true));
+                throw new PrinterException($result);
+            }
+            $jobId = intval($matches[1]);
+        } catch (\Exception $e) {
+            Log::error("Printing error at line: " . __FILE__ . ":" . __LINE__ . " (in function " . __FUNCTION__ . "). " . $e->getMessage());
+            throw new PrinterException($e->getMessage(), $e->getCode(), $e->getPrevious());
```
kdmnk, line 108: I see now. Just return null or -1, the new exception with the same data won't give us value.

app/Models/Printer.php
```diff
@@ -0,0 +1,154 @@
...
+     * @param int $copies 
+     * @param string $path 
+     * @return int The `jobId` belonging to the printjob
+     * @throws PrinterException If the printing fails
+     */
+    public function print(bool $twoSided, int $copies, string $path) {
+        if (config('app.debug')) {
+            return -1;
+        }
+        $jobId = null;
```
kdmnk, line 84: move the return statement inside the try-catch and remove this.

app/Models/PrintAccount.php
```diff
@@ -54,31 +57,101 @@ class PrintAccount extends Model
...
+     * @return mixed
+     * @throws BindingResolutionException
+     * @throws NotFoundExceptionInterface
+     * @throws ContainerExceptionInterface
+     */
+    public static function getBalanceNeeded(int $pages, int $copies, bool $twoSided)
+    {
+        $pageTypesNeeded = self::getPageTypesNeeded($pages, $twoSided);
+
+        return $pageTypesNeeded['one_sided'] * config('print.one_sided_cost') * $copies +
```
kdmnk: Pretty nice, but do we really want to charge separately for the last page in a two-sided document? Without that, the getPageTypesNeeded would be unnecessary and even the Printer::getDocumentPageNumber could just return the actual pages to be printed in total, directly, based on $twoSided. It would really simplify things without barely anyone noticing or complaining.

app/Models/Printer.php
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * Returns the completed printjobs for this printer.
+     * @return void 
+     * @throws NotFoundExceptionInterface 
+     * @throws ContainerExceptionInterface 
+     */
+    public function getCompletedPrintJobs() {
+        try {
+            $process = new Process(['lpstat', '-W', 'completed', '-o', $this->name, '-h', "$this->ip:$this->port", '|', 'awk', "'{print $1}'"]);
```
kdmnk: Let's create our own Process class that inherits this. Add a parameter to 'getOutput' with a default value returned when we are in debug mode. Override run() accordingly.

app/Models/Printer.php
```diff
@@ -0,0 +1,154 @@
...
+
+        return $jobId;
+    }
+
+    /**
+     * Returns the number of pages in the PDF document at the given path.
+     * @param string $path 
+     * @return int 
+     */
+    public static function getDocumentPageNumber(string $path): int {
```
kdmnk: This should be elsewhere. Does it make sense to create a Document class, storing the path, cost and other attributes?

app/Models/Printer.php
```diff
@@ -0,0 +1,154 @@
...
+    public function printJobs() {
+        return $this->hasMany(PrintJob::class);
+    }
+
+    /**
+     * Attemts to cancel the given `PrintJob`. Returns wether it was successful.
+     * @param PrintJob $printJob 
+     * @return PrinterCancelResult
+     */
+    public function cancelPrintJob(PrintJob $printJob) {
```
kdmnk: this should be in the printJob class

app/Models/Printer.php
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * Asks the printer to print a document with the given configuration.
+     * @param bool $twoSided 
+     * @param int $copies 
+     * @param string $path 
+     * @return int The `jobId` belonging to the printjob
+     * @throws PrinterException If the printing fails
+     */
+    public function print(bool $twoSided, int $copies, string $path) {
```
kdmnk: This should be the constructor of the PrintJob class, shouldn't be?

app/Models/PrintJob.php
```diff
@@ -37,46 +41,79 @@ class PrintJob extends Model
...
+     */
+    public function getTranslatedStateAttribute()
+    {
+        return __("print." . strtoupper($this->state->value));
+    }
+
+    /**
+     * Gets the printjob-status with every printer, updates the status of the completed printjobs.
+     */
+    public static function checkAndUpdateStatuses()
```
kdmnk: This should be a method of the Printer, updating it's own printJobs. The loop can be outside the call.

app/Models/Printer.php
kdmnk: Move getBalanceNeeded and similar functions also there.

kdmnk (changes requested):
Nice job, thanks for taking care of this. Please take a look at the issues. Overall, the functions are still a bit too complex, let's revise this again after you fix these. And we need extensive unit tests here (calculating price, pages, etc), otherwise this would be worthless. Let me know if you have questions.

app/Console/Commands/CreateDefaultPrinter.php
sszajbely: This is basically an initialization command, so that inserting a printer into the db by hand is not needed. I can add customization if you think it's necessary.

app/Http/Controllers/Dormitory/Printing/PrintAccountController.php
sszajbely: I think the main point is that it would be a path param, not a request param. But still if you can tell me how I can do that in the blade, that would be nice.

app/Http/Controllers/Dormitory/Printing/PrintAccountController.php
sszajbely: I pretty much kept the functionality of the past code, so this is what was here before, just refactored.

app/Console/Commands/CreateDefaultPrinter.php
kdmnk: But when do you call it? Can't you just create it in a migration then?

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
sszajbely: Because I didn't know if it is used everywhere or not :D

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
sszajbely: Mostly because I didn't want to always have the printer as a dependency for the blade, but I guess it's not really worth much.

app/Models/PrintAccount.php
sszajbely: I don't know, if you think it's worth it being simpler, I can live with that. :D

app/Models/Printer.php
sszajbely: You mean a utility class, not a model, right?

app/Http/Controllers/Dormitory/Printing/PrintAccountController.php
kdmnk: Ah, I see. I have an example how we handle that in the tabulator, with javascript: 
```
.ajax({
                    type: "POST",
                    url: "{{ route('internet.internet_accesses.extend', [':id']) }}".replace(':id', id),
                    ...})
```
You need some js to override the default form sending logic and replace the URL, but I think it might worth it here

app/Http/Controllers/Dormitory/Printing/PrintAccountController.php
kdmnk: Then, if you want, open a new issue.

app/Models/Printer.php
sszajbely: The `PrintJob` should have a `cancel()` function, which calls this function. As the `Printer` is the one cancelling the `PrintJob`, but the `PrintJob` is being cancelled. Right?

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
kdmnk: We only use it in AdminInternetController and  MacAddressController elsewhere. Check them please, or you can open a issue.

app/Http/Controllers/Dormitory/Printing/PrintAccountController.php
horcsinbalint: Is $otherAccount and $printAccount the same PHP object? I'm not sure. If they are different, decreasing $printAccount's balance and saving it, and then increasing $otherAccount's balance and saving it might add money to the account in the end.
Like:
```
a = 1500
b = a #1500
c = a #1500
amount = 300 
b -= amount #1200
a = b #1200
c += amount #1800
a = c #1800
```

app/Http/Controllers/Dormitory/Printing/PrintAccountController.php
```diff
@@ -0,0 +1,128 @@
...
+
+class PrintAccountController extends Controller
+{
+    /**
+     * Updates balance of a `PrintAccount`.
+     * This can be done in two ways: either by topping up the account by giving money to an admin
+     * or by transfering money from one account to the other.
+     */
+    public function update(Request $request)
+    {
```
horcsinbalint, line 29: Shouldn't it be atomic? Like some DB transaction or something like that?

kdmnk:
@machiato32 do you have any update on this?

sszajbely:
> @machiato32 do you have any update on this?

Not yet, sorry.

kdmnk:
Now I had a new idea for the admin/not admin controller separation. 
We merge the two pages into one (originally I did not want to do that but now it suddenly makes sense :D ).
In the backend, we can check based on policies what to query. On the UI we also check the policies if we include the "user" column and other advanced filters for admins or not. 
However, we set a default filter on the current user so no data of other users are shown initially.
The modifier forms can be added next to their respective tables.
What do you think?

sszajbely:
> Now I had a new idea for the admin/not admin controller separation. We merge the two pages into one (originally I did not want to do that but now it suddenly makes sense :D ). In the backend, we can check based on policies what to query. On the UI we also check the policies if we include the "user" column and other advanced filters for admins or not. However, we set a default filter on the current user so no data of other users are shown initially. The modifier forms can be added next to their respective tables. What do you think?

I honestly think this seems like more effort than it's worth. I'll just split the controllers and that should be easy to understand for everyone.

app/Models/Printer.php
sszajbely: I thought about doing it like that, but I think we would need to alter the db structure a bit. Because a PrintJob not only saves the printjob itself, but also its cost etc. So I think it would be a bit strange to put all of that cost calculation and print account modification into the Printer model aswell.

sszajbely:
@kdmnk I tried to resolve everything you mentioned, and also found some bugs.

I still couldn't really test this, and unit/feature tests are also missing, but I think there are a lot of improvements already.

app/Policies/PrintAccountPolicy.php
```diff
@@ -43,4 +40,14 @@ public function modify(User $user): bool
...
+    /**
+     * Determine whether the user can transfer balance from the print account.
+     * @param User $user
+     * @param PrintAccount $printAccount
+     * @return bool
+     */
+    public function transferBalance(User $user, PrintAccount $printAccount): bool
+    {
+        return $user->id == $printAccount->user_id;
+    }
```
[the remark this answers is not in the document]
kdmnk: yes

app/Http/Controllers/Dormitory/Printing/PrintAccountHistoryController.php
kdmnk: @machiato32

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
kdmnk: @machiato32

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
kdmnk: This is still valid. Like what you did in FreePagesController to keep it consistent.

.vscode/launch.json
```diff
@@ -0,0 +1,48 @@
+{
```
kdmnk: do we want these files in git?

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
kdmnk: Ah I see, I'm stupid, sorry

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
kdmnk: Still relevant. Modifying free pages or balance should be at least in a helper function. Also the document should be printed after we modified the balance as it can not be rolled back on fail.

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
kdmnk: Or name it update but add another parameter what should be the new status. Maybe later we can reimburse prints or whatever.

app/Models/PrintJob.php
```diff
@@ -37,46 +42,92 @@ class PrintJob extends Model
...
+    /**
+     * Attemts to cancel the given `PrintJob`. Returns wether it was successful.
+     * @param PrintJob $this
+     * @return PrinterCancelResult
+     */
+    public function cancel()
     {
-        return function ($data) {
-            $data->cost = "{$data->cost} HUF";
+        $printer = $this->printer ?? Printer::firstWhere('name', config('print.printer_name'));
```
kdmnk: Again, why don't we always have a printer set in the db? I would delete the config entry or use only that to avoid confusion.

routes/web.php
```diff
@@ -111,22 +114,20 @@
...
-    Route::post('/print/transfer_balance', [PrintController::class, 'transferBalance'])->name('print.transfer-balance');
-    Route::post('/print/print_jobs/{id}/cancel', [PrintController::class, 'cancelPrintJob'])->name('print.print_jobs.cancel');
-    Route::put('/print/print', [PrintController::class, 'print'])->name('print.print');
-    Route::middleware(['can:modify,App\Models\PrintAccount'])->group(function () {
-        Route::get('/print/account_history', [PrintController::class, 'listPrintAccountHistory'])->name('print.account_history');
-        Route::get('/print/manage', [PrintController::class, 'admin'])->name('print.manage');
-        Route::post('/print/modify_balance', [PrintController::class, 'modifyBalance'])->name('print.modify');
-    });
-    Route::post('/print/add_free_pages', [PrintController::class, 'addFreePages'])->name('print.free_pages')->middleware('can:create,App\Models\FreePages');
+    Route::get('/printer/{page?}', [PrinterController::class, 'index'])->name('printer.index'); // also print.manage, print
```
kdmnk: what are the stuff in comments?

kdmnk (changes requested):
Looks much cleaner but some comments still applies. Also take a look at the coderabbit comments which are pretty cool.
We can take a look together if you want.

kdmnk (approved):
The code looks nice now, good job and thanks!
Let's test this.

sszajbely:
I would like to hand over the testing phase of this PR to someone who actually still lives in the building, especially since I cannot test the printing itself, which is the most important part of this. @EotvosCollegium/marscodeowners.

app/Console/Commands.php
```diff
@@ -15,65 +16,15 @@ private static function isDebugMode()
...
-
     public static function pingRouter($router)
     {
         if (self::isDebugMode()) {
             $result = rand(1, 10) > 9 ? "error" : '';
         } else {
             // This happens too often to log.
-            $command = config('commands.ping') . " " . $router->ip . " -c 1 | grep 'error\|unreachable'";
-            $result = exec($command);
+            $process = new Process(['ping', $router->ip, '-c', '1', '|', 'grep', "'error\|unreachable'"]);
```
horcsinbalint: I have no experience with `Symfony\Component\Process\Process`, but in other languages similar concepts of processes would not allow multiple commands being piped together.

app/Console/Commands.php
```diff
@@ -83,10 +34,9 @@ public static function latexToPdf($path, $outputDir)
         if (self::isDebugMode()) {
             $result = "ok";
         } else {
-            $command = config('commands.pdflatex') . " " . "-interaction=nonstopmode -output-dir " . $outputDir . " " . $path . " 2>&1";
-            Log::info($command);
-            $result = exec($command);
-            Log::info($result);
+            $process = new Process(['pdflatex', '-interaction=nonstopmode', '-output-dir', $outputDir, $path]);
```
horcsinbalint: As far as I remember, the `pdflatex` command did not work inside the container in the case of the php interpreter, that is why `config('command.pdlatex')` was used.

sszajbely:
The only test failing now is the `PrintControllerTest` which I haven't updated yet. So I think all major concerns asides from tests are being resolved now, so as I stated earlier, I would like to hand off this part to someone in the building @EotvosCollegium/marscodeowners

app/Http/Controllers/Dormitory/Printing/PrintJobController.php
```diff
@@ -0,0 +1,185 @@
...
+
+namespace App\Http\Controllers\Dormitory\Printing;
+
+use App\Enums\PrintJobStatus;
+use App\Http\Controllers\Controller;
+use App\Models\FreePages;
+use App\Models\PrintAccount;
+use App\Models\Printer;
+use App\Enums\PrinterCancelResult;
+use App\Models\PrinterHelper;
```
horcsinbalint: `App\Models\PrinterHelper` does not exists, you probably meant `App\Utils\PrinterHelper`

sszajbely: Good catch, namespace was wrong.

resources/views/user/printing.blade.php
```diff
@@ -11,14 +11,14 @@
                     @foreach($user->freePages->sortBy('deadline') as $page)
                         <li>
                         <span class="new badge
-                            @if($page->available())
+                            @if($page->available)
```
horcsinbalint, line 14: `available` is a function, not an attribute. It should be `available()`

app/Models/FreePages.php
```diff
@@ -51,23 +53,73 @@ class FreePages extends Model
...
     {
-        return $this->belongsTo('App\Models\PrintAccount', 'user_id', 'user_id');
+        return $this->belongsTo(PrintAccount::class, 'user_id', 'user_id');
+    }
+
+    /**
+     * Wether the free pages are still available.
+     * @return bool
+     */
+    protected function available()
```
horcsinbalint: This fucntion is called outside from `printing.blade.php`, should not be protected, it should be public.

resources/views/user/printing.blade.php
sszajbely: It is an attribute though.
```
protected function available()
    {
        return Attribute::make(
            get: fn () => now()->isBefore($this->deadline)
        );
    }
```

database/factories/PrintJobFactory.php
```diff
@@ -13,8 +14,7 @@ public function definition()
     {
         return [
```
horcsinbalint, line 15: PrintJob requires a `printer_id` attribute to be added during seeding as there is no default value.

app/Models/FreePages.php
sszajbely: AFAIK attributes don't care about the visibility of the function behind it. Virtually every attribute in the project is created in a protected function.

horcsinbalint: For me route `/profile` was crashing with error "App\Models\FreePages::available must return a relationship instance.". The return value type should be included according to the documentation (https://laravel.com/docs/11.x/eloquent-mutators#attribute-casting): "All attribute accessor / mutator methods must declare a return type-hint of `Illuminate\Database\Eloquent\Casts\Attribute`":
```php
protected function available() : Attribute
```

horcsinbalint: With this change, it works fine on my machine

sszajbely: Ah so it needs the type hint, okay. That's why I don't like use the new syntax for attribute creation @kdmnk :P. But thanks, will fix it tomorrow

horcsinbalint: Thanks, there are also other instances where it should be fixed (like `App\Models\PrintJob::translatedCost`).

app/Console/Commands.php
```diff
@@ -10,84 +11,23 @@
...
-
     public static function pingRouter($router)
     {
-        if (self::isDebugMode()) {
-            $result = rand(1, 10) > 9 ? "error" : '';
-        } else {
-            // This happens too often to log.
-            $command = config('commands.ping') . " " . $router->ip . " -c 1 | grep 'error\|unreachable'";
-            $result = exec($command);
+        if (preg_match('/[a-zA-Z]/', $router->ip)) {
```
horcsinbalint: Nitpick, as this is already an improvement: could we check if it is in fact an IP address, and if not, then throw an exception? Like using this regex? `^(?:(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9][0-9]|[0-9])\.){3}(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9][0-9]|[0-9])$`
Some quite painful shellscript could be injected here even if they do not control any letters. (Like a fork bomb: `:(){ :|:& };:`)

horcsinbalint:
I added a few proposed changes in #669

horcsinbalint:
I will decrease the scope of #669 , and merge them and try it out in production on Saturday if no other issues are raised

horcsinbalint:
I gave up on reviewing all of the changes. I fixed a few bugs.
I am merging this for now, but I will revert the PR if anything goes wrong.

horcsinbalint:
There was an issue, reverted, a new PR will be soon opened

Issue #364, opened 2023-11-25
Caching is not properly supported


horcsinbalint:
Caching in a production server can be important using Laravel. By running the following code to make caches, there are issues in blades that `{{ENV('EXAMPLE')}}` are not loading on the page.
```
php artisan event:cache
php artisan route:cache
php artisan view:cache
php artisan config:cache
```

kdmnk:
I don't understand. We did not use artisan caching so far.

horcsinbalint:
To improve the security, maintainability and reliability of our systems I'm doing some restructuring of our servers. Since they might come with some performance burden and I would like to meet the recommendations set forth in the Laravel documentation (https://laravel.com/docs/9.x/configuration#configuration-caching, https://laravel.com/docs/9.x/deployment#optimizing-configuration-loading), I would also would like to set up proper caching for used resources.

The changes that were done in the [deployment](https://github.com/EotvosCollegium/mars/tree/deployment) branch seem to be enough to allow proper caching.

I'm planning on making proper branching and PRs for the upstream [development](https://github.com/EotvosCollegium/mars/tree/development) branch, but there were several issues during the migration of the infrastructure and I was not yet able to properly document all of the features that did not work as expected.

Issue #373, opened 2023-12-19
Pdflatex command was not found


horcsinbalint:
When serving Mars's dynamic content from a Docker container running PHP-FPM, we had an issue with the `pdflatex` command not being found. The necessary packages were installed in the container and the command given to the `exec` function in `app/Console/Commands.php` were successful if I executed it inside the container in a `/bin/bash` terminal. Probably there is an issue that the PHP-FPM and bash has different `$PATH` values. As a quick fix, 5dc985ea3b16d1ca9d1cd8fe3fea895735432d18 contains `pdflatex`'s full path.

Most probably it is an issue with our deployment (giving PHP-FPM a wrong path), so there is probably not an issue with Mars. Potentially setting the commands in .env could fix the issue, but not setting a proper path could be attributed to user error.
So this issue is most probably invalid.

kdmnk:
I found these issues (with #372, #366) slightly irrelevant as these should be a part of your next PR. Or do you want to have the deployment branch co-existing with development in the future? Because then we need to reconfigure other stuff as well.

For adding stuff into .env, just add it to .env.example and you can create a new config file for the server related stuff where you map them.

horcsinbalint:
I added these issues so they would not be forgotten and I would like to include these changes in my upcoming PR, which would close these issues. In the past with other (mostly closed-source) projects we firstly created an issue for new features, and then developed it, and later created a PR, which closed these issues. (I might be mistaken, probably this is not how this project works.)

For the deployment and development branch: I would like them to co-exist. The deployment branch is the branch that is actually deployed on the server (in each deployment, that branch is cloned from the repo) and sometimes hotfixes are necessary that might be only compatible with our own environment (like #372). In this way, every change is version controlled, and it is not possible to make hot-fixes on the server itself. (So it takes time for changes to propagate. Which is both a bug and a feature: during this time somebody might realize a mistake.)
So I think new features and most changes should be done firstly in seperate branch, then after a PR, merged to the development branch, and then brought to the deployment branch. And the deployment branch can be used to make hotfixes that are not compatible with different deployments.

So I think the development branch should have a relatively clean history with properly managed PRs, and a deployment branch that might have some hotfixes that are necessary for the operation of our own deployment. But these divergencies should always be removed after some time.

It is probably a bad idea, and I think we should start a discussion about whether we should do this in a different manner. I don't know how these are usually done.
Best would be to always use properly written code and always deploy a stable branch but I think sometimes we do need temporary fixes.

[...]

Issue #372, opened 2023-12-19
Make TrustProxies configurable from .env


horcsinbalint:
We have put Mars behind a reverse proxy. This requires Laravel to trust the given proxy according to the [documentation on Trusted Proxies](https://laravel.com/docs/9.x/requests#configuring-trusted-proxies). There has been a temporary fix applied in 6053d95c47e4f949460bfe1110293695b44a06d0 which has a hard-coded IP set for the proxy. It would be beneficial to be able to set this value in the .env file.

PR #396, opened 2024-01-05
Pulling back new changes from deployment

viktorcsimma:
This is to merge the improvements @horcsinbalint has made to the development branch.

@kdmnk what do you think? ~~I can review it, but maybe only on Sunday or after that.~~ Oh, the system doesn't let me do it; since it's technically my PR... but I can add comments, so it could be solved.


app/Console/Commands.php
```diff
@@ -83,9 +83,10 @@ public static function latexToPdf($path, $outputDir)
         if (self::isDebugMode()) {
             $result = "ok";
         } else {
-            $command = "pdflatex " . "-interaction=nonstopmode -output-dir " . $outputDir . " " . $path . " 2>&1";
+            $command = "/usr/bin/pdflatex " . "-interaction=nonstopmode -output-dir " . $outputDir . " " . $path . " 2>&1";
```
kdmnk: add the full path to the config then

app/Http/Middleware/TrustProxies.php
```diff
@@ -12,7 +12,7 @@ class TrustProxies extends Middleware
      *
      * @var array|string
      */
-    protected $proxies;
+    protected $proxies = ['172.5.0.10'];
```
kdmnk: What's this and why?

kdmnk (changes requested):

app/Http/Middleware/TrustProxies.php
horcsinbalint: We are running mars behind a reverse proxy and setting the reverse proxy as a trusted proxy for Laravel is mandatory if we want to use HTTPS. (https://laravel.com/docs/9.x/requests#configuring-trusted-proxies) There are some other hacky ways of setting it up but I didn't like them.
I would like it to be configurable from the .env, and I will add this to the PR.

kdmnk: Okay, and add a comment here. Will this affect local development?

horcsinbalint:
Unfortunately I'm currently quite busy with my other tasks and other issues with the server. I'm planning on fixing the issues with this PR in about 3-4 weeks.

app/Http/Middleware/TrustProxies.php
horcsinbalint: I'm not 100% sure, and have not tested it, but I think it only comes in to effect if at least one of the headers in the request:
```
Request::HEADER_X_FORWARDED_FOR |
Request::HEADER_X_FORWARDED_HOST |
Request::HEADER_X_FORWARDED_PORT |
Request::HEADER_X_FORWARDED_PROTO |
Request::HEADER_X_FORWARDED_AWS_ELB
```
Which are mostly stuff that can come from a reverse proxy / load-balancer. But I think it should not have any effect unless you are connecting from IP `172.5.0.10` and the `APP_URL`'s protocol is https.

I will look into it today or tomorrow, but I think it should not have any effect for local development.

horcsinbalint: I have tested it in my local environment and have not found any issues. So I think it should be fine

.env.example
```diff
@@ -68,6 +68,8 @@ PREVENT_ACCESSING_MISSING_ATTRIBUTES=false
 PRINT_COST_ONESIDED=8
 PRINT_COST_TWOSIDED=12
 PRINTER_NAME=NemUjBela
+PRINTER_ADDITIONAL_ARGS=
+PRINTER_STAT_ADDITIONAL_ARGS=
```
BertalanD, line 72: I'm not sure how much of the changes to printing will be necessary once the printer server is redesigned (eotvoscollegium/smaugreinstall#8). I'm planning on tackling that task on Thursday; could we delay this part until then?

app/Http/Middleware/TrustProxies.php
BertalanD: Could we make this configurable instead of hard-coding the IP in the source code?

horcsinbalint: Yep, I would like to make it configurable, I just currently have a long lead-time

horcsinbalint: Or if someone would fix it, I would happily give advice and / or check the code

horcsinbalint:
I think I have fixed the issues that have been pointed out. @kdmnk @viktorcsimma Could you please check it?

Should fix #364 #366 #372 #373

kdmnk (approved):

PR #404, opened 2024-01-21
Fix error with faults page

horcsinbalint:


BertalanD:
Was this the only instance where the path to the internet settings wasn't updated in #386?

BertalanD (approved):

horcsinbalint:
> Was this the only instance where the path to the internet settings wasn't updated in #386?

I have searched for `"internet"` and `'internet'` and this was the only occurrence. But there could be similar issues, I have not checked it yet.

PR #408, opened 2024-01-21
Set up code owners correctly

horcsinbalint:
So that not everybody with write permissions can approve PR-s

### DIFF
--- a/.github/.CODEOWNERS
+++ b/.github/.CODEOWNERS
@@ -1,1 +1,0 @@
-* @EotvosCollegium/sysadmin
--- a/.github/.CODEOWNERS
+++ b/.github/.CODEOWNERS
@@ -1,0 +1,1 @@
+* @EotvosCollegium/sysadmin
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @EotvosCollegium/marscodeowners


kdmnk:
How does this work?

horcsinbalint:
> How does this work?

If I understand correctly, it has been set that only Code Owners can accept PRs. By default it probably means that anybody with write permission to the repo can approve PRs (I'm not 100% sure on that though). If we set up the CODEOWNERS file, we can manage who can approve and what files.
In this case, I have set that every file is owned by the members of the EotvosCollegium/marscodeowners group can accept any files. (I did not want to ping everybody, here you can check the members: https://github.com/orgs/EotvosCollegium/teams/marscodeowners)

horcsinbalint:
Sorry, forgot a minute ago, that I have checked before creating this PR that none of the files had owners. So I'm 95% sure that we need to set up this file to protect the repo. (Now none of the files has a matching rule, so it can be approved by anybody - most probably needs write permission though.)

kdmnk:
Okay, nice but please double-check. With your recent PR (https://github.com/EotvosCollegium/mars/pull/415), if that works, I don't think we need to give write permission to non sysadmins at all (aka give them EotvosCollegium membership on GitHub). Any others can have read access and only push through PRs from forks, and give them rights on hidden repos respectively and separately.

This just feels a little hacky and I feel we are quite a small team to rely on these solutions. But if no, I'm fine with it

horcsinbalint:
> Okay, nice but please double-check

I have checked on a different repo, and it works as I described.

> With your recent PR (https://github.com/EotvosCollegium/mars/pull/415), if that works, I don't think we need to give write permission to non sysadmins at all (aka give them EotvosCollegium membership on GitHub). Any others can have read access and only push through PRs from forks, and give them rights on hidden repos respectively and separately.

Yeah, that would be the ideal workflow, and I would like it more, but in the discussion on Discord it was decided that we would give out write permissions to students that take on the course. I am not a big fan of it, but by protecting our important branches, we should be able to make it relatively safe (except for [accidentally] pushing large amounts of data on any of the branches, which would make cloning the repo a bit of a pain, since it would be in the git history).

I only had two ideas to protect our important branches (other than everyone creating their own forks). One was to set up Code Owners (this PR) and the other one was to create a fork from the repo, and give write access to that fork to the students. In this case, they would not need to create their own forks, and they would still be able to contribute to this project via PRs.

I think @viktorcsimma knows more on the reasoning of giving write permission to the students.

viktorcsimma:
The original reason was that forks might be a little confusing at first. So even when I work with such repos, I often forget to update my fork, and then I don't understand what is going on.

But actually, @kdmnk you have convinced me. It might even be better to force them to use forks; we tell them to always have an up-to-date fork. If they forget it and make mistakes, they will just remember it more:D

horcsinbalint:
I would like to close this pull request either by merging it to the correct branch, or by closing it. I don't have a clear opinion on which would be the better course of action. I think it would not cause any harm to set up the code owners, but also would not make a big difference. It would automatically ask for review from code owners, which would probably be nice.

kdmnk (approved):

PR #412, opened 2024-01-25
Add access management for calculating workshop balance

horcsinbalint:
Should fix #411

PR #413, opened 2024-01-25
Add access management for calculating workshop balance (#412)

horcsinbalint:
* Add access management for calculating workshop balance

* style: format code with PHP CS Fixer

This commit fixes the style issues introduced in cb182ca according to the output from PHP CS Fixer.

Details: https://github.com/EotvosCollegium/mars/pull/412

* Change to Checkout::class

---------


horcsinbalint:
Removed `Require branches to be up to date before merging` and `Require linear history` for branch `deployment`. The first one sometimes could be difficult to meet. The second one should not be used, since `development` contains lots of squashed commits.

PR #428, opened 2024-02-14
XAMPP installation guide for Windows

viktorcsimma:
This worked for me without mysterious errors; I think I would even recommend it to students.

### DIFF
--- a/_SETUP_DEV_ENVIRONMENT.md
+++ b/_SETUP_DEV_ENVIRONMENT.md
@@ -22,6 +22,63 @@ There are several ways to set up your development server and environment. Basica
 
 Note: to regenerate the docker configuration, use `php artisan sail:install --devcontainer`
 
+### Windows (natively)
+On Windows, probably the optimal option with respect to performance is
+to simply install a web server and a database server natively on the system,
+instead of using containers or virtual machines.
+However, the installation itself is a bit more complicated.
+(But actually, there aren't as many mysterious errors as with containers.)
+
+1. Download and install the following programs:
+    - [XAMPP](https://www.apachefriends.org/download.html) (the newest available one) – this includes the PHP interpreter, the Apache web server and the MySQL database server.
+        - Besides the obligatory options, we only need MySQL, phpmyadmin and Fake Sendmail.
+        - You don't need to allow it through the firewall, as we are only going to access these locally.
+    - [Composer](https://getcomposer.org/download/) – this is a package manager for PHP.
+        - It's enough to install it only for your account when asked.
+    - [NVM](https://github.com/coreybutler/nvm-windows/releases) (choose `nvm-setup.exe`) – this downloads Node.js and the NPM package manager, which will again download other things:D
+    - [Git](https://gitforwindows.org/) – the version manager.
+2. Open `C:\xampp\php\php.ini`. This is PHP's configuration file. We need to enable two extensions here for things to work.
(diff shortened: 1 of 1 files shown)


_SETUP_DEV_ENVIRONMENT.md
```diff
@@ -22,6 +22,64 @@ There are several ways to set up your development server and environment. Basica
 
 Note: to regenerate the docker configuration, use `php artisan sail:install --devcontainer`
 
+### Windows (natively)
+On Windows, probably the optimal option with respect to performance is
+to simply install a web server and a database server natively on the system,
```
kdmnk, line 27: This is not natively, it's based on Xampp, which is a developer tool and also does a lot of "magic" in the background. We had issues with Xampp before, which did not support the newest PHP versions. Apparently, I see that currently this is not the case, however, we should make a note here to check it before even trying to set up.

_SETUP_DEV_ENVIRONMENT.md
```diff
@@ -22,6 +22,64 @@ There are several ways to set up your development server and environment. Basica
...
+    - [Git](https://gitforwindows.org/) – the version manager.
+        - For the text editor, the simplest is to choose Windows' Notepad (or VS Code, if you have it). Otherwise, you can stick to the defaults.
+2. Open `C:\xampp\php\php.ini`. This is PHP's configuration file. We need to enable two extensions here for things to work.
+    - To do this, search for the lines `;extension=zip` and `;extension=gd`, and delete the semicolon (`;`) before them. (If there are more than one copies of them, it's enough to enable only one instance.)
+3. Open a command prompt (open Start menu and type `cmd`).
+4. Type `nvm install lts`. This command will install NPM for you.
+5. To be able to use it, you need to add the folder containing the NPM binary to the operating system's PATH variable (so that the command prompt will find it).
+    - Check the installed version of Node.js with `nvm list`
+    - Follow [these steps](https://www.mathworks.com/matlabcentral/answers/94933-how-do-i-edit-my-system-path-in-windows). Add the path `%USERPROFILE%\AppData\Roaming\nvm\vXX.XX.X`, where `XX.XX.X` is the version printed by `nvm list`.
```
kdmnk, line 47: Avoid links to random pages which might not be available/updated later. Just mention adding it to the path, everyone can google it.

_SETUP_DEV_ENVIRONMENT.md
```diff
@@ -22,6 +22,64 @@ There are several ways to set up your development server and environment. Basica
...
+    - Select `User accounts` and add a new user. This is what Mars is going to use. Name it `mars` and give it the password you wrote in `.env`.
+    - After you created it, click `Edit privileges` and switch to the `Database` tab. Click on the `mars` database you created in the first step, and then `Go`. `Check all` and `Go` again.
+    - You can now close this page.
+14. To initialise the database (with fancy test data!) and some other things, switch back to the command prompt and while being in the `mars` folder, run:
+```bat
+php artisan key:generate
+php artisan migrate:fresh --seed
+```
+15. Finally, you can start your server! For this, type `php artisan serve`. (To stop the server, press Control-C. If you want to use the command prompt, you will need to open a new window.)
+16. Your local site can be accessed from the browser at `127.0.0.1:8000`. Log in with `example@eotvos.elte.hu` and `asdasdasd`.
```
kdmnk, line 75: The credentials are coming from .env just reference that.

kdmnk (changes requested):

PR #432, opened 2024-02-19
Obey browser language preferences

BertalanD:
Before this change, the site would always be displayed in Hungarian until the user explicitly set their language preferences in the menu. This is inconvenient for guests who do not speak the language; I personally had to had to help a guest sign up, as they did not notice the hamburger menu on their phone.

This change makes the `Locale` middleware select the appropriate language via the `Accept-Language` header. It can still be overridden via the Language menu.

### DIFF
--- a/app/Http/Middleware/Locale.php
+++ b/app/Http/Middleware/Locale.php
@@ -16,7 +16,8 @@ class Locale
      */
     public function handle($request, Closure $next)
     {
-        App::setLocale($request->cookie('locale', config('app.locale')));
+        $locales = array_keys(config('app.locales'));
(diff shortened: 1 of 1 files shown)


kdmnk (approved):
Nice. Have you checked if all our language keys match the possible values in the cookie?

BertalanD:
Honestly, I've only checked English and Hungarian; I've never seen *any* of the other translations show anything besides the fallback English strings (even before this change).  Will take a look later.

BertalanD:
It looks like the language selector component also gets the codes from the keys of `app.languages`, so this should be fine.

PR #446, opened 2024-02-28
Instructions for docker-based dev setup

Trigary:
This setup is just a more detailed version of @horcsinbalint's instructions.

### DIFF
--- a/docker-dev-setup/Dockerfile
+++ b/docker-dev-setup/Dockerfile
@@ -12,7 +12,7 @@ RUN apt-get update \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 
 # Install Node.js
-RUN wget -O node.tar.xz https://nodejs.org/dist/v20.9.0/node-v20.9.0-linux-x64.tar.xz && \
+RUN wget -nv -O node.tar.xz https://nodejs.org/dist/v20.9.0/node-v20.9.0-linux-x64.tar.xz && \
     tar -xf node.tar.xz && \
     rm node.tar.xz
 ENV PATH=$PATH:/workspace/node-v20.9.0-linux-x64/bin
--- a/docker-dev-setup/Dockerfile
+++ b/docker-dev-setup/Dockerfile
@@ -3,12 +3,13 @@ FROM ubuntu:23.10
 WORKDIR /workspace
 
 # Install PHP and other dependencies
-RUN apt update \
-    && apt install -y wget tar xz-utils iputils-ping composer poppler-utils cups-client cm-super \
-                      libzip-dev libpng-dev libonig-dev libcurl4-openssl-dev libxml2-dev \
-                      php8.2-cli php8.2-mysql php8.2-zip php8.2-gd php8.2-mbstring php8.2-curl php8.2-xml php8.2-bcmath \
-                      texlive-latex-base texlive-latex-extra texlive-lang-european \
-    && apt clean && rm -rf /var/lib/apt/lists/*
+RUN apt-get update \
+    && apt-get upgrade -y \
+    && apt-get install -y wget tar xz-utils iputils-ping composer poppler-utils cups-client cm-super \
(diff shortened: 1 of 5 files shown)


horcsinbalint:
Thanks for making documentation for it and making it more understandable overall. I might be biased but I would recommend mine more than the "Universal method", as both require Docker but mine does not need composer and php on the host system. So I would probably put it above that.

docker-dev-setup/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+This setup uses Docker to run the app and the database,
+therefore it is very easy to set up on any platform that supports Docker (Linux, Windows via WSL2, etc.).
+This setup is also recommended if you don't want to install PHP and other dependencies on your host machine,
+or if you already have a different (incompatible) version of PHP installed.
+
+## Requirements
+
+- [Docker](https://docs.docker.com/engine/)
+- [Docker Compose](https://docs.docker.com/compose/)
```
horcsinbalint: Is Docker Compose a requirement? AFAIK newer Docker Engine versions come bundled with the `docker compose` command by default. If it is not available, they should update their Docker installation,

docker-dev-setup/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ubuntu:23.10
+
+WORKDIR /workspace
+
+# Install PHP and other dependencies
+RUN apt update \
```
horcsinbalint: Use `apt-get` instead of `apt` in Dockerfiles as `apt` does not have a stable interface

docker-dev-setup/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+    && apt install -y wget tar xz-utils iputils-ping composer poppler-utils cups-client cm-super \
```
horcsinbalint: `apt`->`apt-get`

docker-dev-setup/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+                      libzip-dev libpng-dev libonig-dev libcurl4-openssl-dev libxml2-dev \
+                      php8.2-cli php8.2-mysql php8.2-zip php8.2-gd php8.2-mbstring php8.2-curl php8.2-xml php8.2-bcmath \
+                      texlive-latex-base texlive-latex-extra texlive-lang-european \
+    && apt clean && rm -rf /var/lib/apt/lists/*
```
horcsinbalint: `apt`->`apt-get`

docker-dev-setup/Dockerfile
horcsinbalint: Please also use `apt-get upgrade -y` to upgrade the packages to the latest version

docker-dev-setup/README.md
horcsinbalint: We can probably say that this requires Docker Engine version 25.0.1 or newer. In that case `docker compose` should most definitely work.

Trigary: Docker Compose is a requirement. But you are right, it doesn't need to be manually installed. What do you suggest?

Trigary:
I have made the requested changes, thank you. I have also placed this new description above the Universal method. Currently, both are marked as recommended, which is obviously not ideal, but I also feel biased in this matter and therefore don't wish to un-recommend the previously recommended solution myself.

docker-dev-setup/README.md
horcsinbalint: I have created a PR for your PR which addresses the issue: https://github.com/Trigary/mars/pull/1

horcsinbalint (commented):
LGTM, but I would still want someone else from @EotvosCollegium/marscodeowners to check it since I wrote the original version of the Dockerfile and docker-compose

kdmnk (approved):
As long as you don't mess up the other methods and it works for you, I don't care :D

Issue #402, opened 2024-01-21
Support Android 11+ / inform user about lack of support for Android 11+


horcsinbalint:
Currently, we use a self-signed certificate for the wifi network.
It is not supported on some phones to turn off certificate checks. (https://www.xda-developers.com/android-11-break-enterprise-wifi-connection/)
We should either publish our CA's cert (it is regenerated every year, so it would not solve the issue - longer timeframe could be problematic), inform users of discontinued support for Android 11+ (many users would be impacted), or use a proper certificate and add a description on how to set up properly (it needs some operational work and probably needs a reconfiguration of all client devices but should not be a big issue).
In either case, we will need to update the description.

horcsinbalint:
#450 fixed this issue

PR #450, opened 2024-03-01
Updating and expanding instructions for connecting to the radius network

viktorcsimma:
I've just helped someone connect to the network and for her, the only way her phone worked was by setting "Use system certificate" and the domain "wifi.eotvos.elte.hu". I've also made other changes and clarifications and included a link to the Windows guide I often use.


horcsinbalint (changes requested):
Do not use external links

viktorcsimma:
Why not? It worked for me a lot of times; we can still update it if it breaks sometime.

horcsinbalint:
Someone might get scammed. No one likes to update because an external website seizes to operate

resources/lang/en/internet.php
horcsinbalint, line 1: The domain name should be configurable in .env

viktorcsimma:
Well, it would be great if someone could write an internal guide for the same; otherwise, I would insist on this. I think the benefits clearly outweigh the security risks.

@kdmnk what do you think?

kdmnk:
If you say so, we can live with that. How often do you experience this specific issue? And yes maybe put that in .env

viktorcsimma:
I think there have been ca. 5 cases where I followed (or simply sent) this tutorial and it helped. There was someone who had not had access for more than a year before I did that.

But putting it into .env is a good idea; I’ll take a look at it in the evening.

horcsinbalint:
Does Windows care about it? I don't know, I usually don't work with such issues, but AFAIK if you simply forget the network and readd it, Windows does not care about what is the CA.
It is also a bit long description for the average user to follow.

horcsinbalint:
Here is a Hungarian description from IIG: https://iig.elte.hu/dstore/document/3505/2020_wifi_installwin10.pdf

horcsinbalint:
Should fix #402

kdmnk:
Can we close this as duplicate to #452?

horcsinbalint:
> Can we close this as duplicate to https://github.com/EotvosCollegium/mars/pull/452?

That was just a PR for the `internet-instructions` branch as it only contained suggestions for this PR.

horcsinbalint (commented):
With the recent changes, I'm no longer blocking the PR. But please @kdmnk check on it before merging

horcsinbalint (approved):

PR #451, opened 2024-03-02
Modularized features

horcsinbalint:
It has been rumored that some other colleges might use this software if it was easier to customize. They might not need some of our features so I made it possible to switch off certain feature sets.

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -27,13 +27,21 @@ class Kernel extends ConsoleKernel
         $schedule->call(function () {
             \App\Models\EventTrigger::listen();
         })->daily()->at('13:00');
-        foreach (\App\Models\Internet\Router::all() as $router) {
-            $schedule->job(new \App\Jobs\PingRouter($router))->everyFiveMinutes();
+        if(Feature::isFeatureEnabled("internet.wireless.routers")){
+            foreach (\App\Models\Internet\Router::all() as $router) {
+                $schedule->job(new \App\Jobs\PingRouter($router))->everyFiveMinutes();
+            }
         }
-        $schedule->job(new \App\Jobs\ProcessWifiConnections())->dailyAt('01:00');
 
-        $schedule->command('backup:clean')->daily()->at('01:00');
-        $schedule->command('backup:run --only-db')->daily()->at('01:30');
+        if(Feature::isFeatureEnabled("internet.wireless.connections")){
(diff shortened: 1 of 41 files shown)


kdmnk:
This does not make any sense to me in our case. As you've also mentioned this is just a rumor. If we really want to put effort into making other organisations use our software, we should start by asking them their concrete needs. Then we can negotiate a method to start working on it. 

Feature flags can be useful but on its own it's not "modularisation" and with this implementation it's easy to leave out a check while it makes the code harder to read which will only make our life harder. 

@machiato32 what do you think? 

I'd recommend to make an issue first and decide what and how do we want to do there, before we start working on it. That also applies to other(/every) changes.

sszajbely:
> This does not make any sense to me in our case. As you've also mentioned this is just a rumor. If we really want to put effort into making other organisations use our software, we should start by asking them their concrete needs. Then we can negotiate a method to start working on it.
> 
> Feature flags can be useful but on its own it's not "modularisation" and with this implementation it's easy to leave out a check while it makes the code harder to read which will only make our life harder.
> 
> @machiato32 what do you think?
> 
> I'd recommend to make an issue first and decide what and how do we want to do there, before we start working on it. That also applies to other(/every) changes.

I totally agree with what you've written, and I want to stress the importance of communication in such cases. This is a big project with many people working on it. So if there is a need/wish for bigger changes (impacting a big part of the codebase) the discussion about if and how the feature is to be implemented is crucial.

And just one comment on the topic of the PR: allowing to turn features on and off is just the first step to a more generally usable administration system, and it is not even the most problematic part. The permissions and roles role system implemented coming from the ruleset of our dormitory is as of now totally hand crafted to our needs: One would need to make a general way of creating roles and managing their permissions, which would go as deep into the code as the policy files. Doing this in a clean and expandable way is really hard on its own not even speaking about the difficulty it would add to the understanding for new developers (mostly 1st year students of the dormitory). And we would do all this just for the hope of some other dormitory maybe using this system, all the while we have much bigger and more pressing issues.

horcsinbalint:
Yes, I agree. I was asked to make a larger feature to mars and it was an option to make it possible to adapt in other Colleges by making some features toggleable, but talk with you before implementing it. I was just fiddling around a bit, and when I had enough information to start a discussion on it, this PR was almost complete. It did not feel like a big change, only a bit of refactoring, and did not take a long time.

I am happy to close the PR without merging, as I also agree that this can make the codebase harder to read and maintain.

PR #452, opened 2024-03-02
Instructions stored on the main webpage

horcsinbalint:
My recommendations for #450


horcsinbalint:
TODO: move into blade so data from .env can be used and would also be more elegant

kdmnk (changes requested):
This should still not be part of our codebase. Put a pdf file with the guide on the server and add a link with a configurable path to it.

routes/web.php
```diff
@@ -136,6 +136,8 @@
             'index', 'store', 'update', 'destroy'
         ]);
 
+        Route::get('/help_windows11', [InternetController::class, 'helpWindows11'])->name('help_windows11');
```
kdmnk: delete if not used

kdmnk (approved):
Just one comment, after that you can merge

PR #456, opened 2024-03-03
Fix issues with escapings with forms

### LINKED ISSUE #28: Fix epistola bugs
See https://github.com/luksan47/mars/issues/639

horcsinbalint:
There have been multiple issues with characters being escaped incorrectly. Check #28

It's better and easier to use textareas if someone needs to show previous user input since it is much more established how one can escape HTML as simple text on the page in PHP / Laravel compared to escaping inside of an attribute.
`x-input.textarea` has been also reworked to have a similar interface to `x-input.text` and also to use slots instead of attributes to show values input by the user (https://laravel.com/docs/10.x/blade#slots).

This PR should fix #28.

### DIFF
--- a/app/View/Components/Input/Textarea.php
+++ b/app/View/Components/Input/Textarea.php
@@ -7,6 +7,7 @@ use App\View\Components\Input;
 class Textarea extends Input
 {
     public $helper;
+    public $withoutLabel;
 
     /**
      * Create a new textarea input instance.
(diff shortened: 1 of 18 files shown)


horcsinbalint:
Fixed conflicts caused by #454

kdmnk (commented):
So after this change when should one use the simple text field? Why don't we delete the simple text field and use only the textarea?

horcsinbalint:
> So after this change when should one use the simple text field? Why don't we delete the simple text field and use only the textarea?

As I understand, currently our `x-input.text` implementation works and used as if it was just a fancy `<input>` tag. We use it for types other than `text` (like `number`). For them, `textarea`s would not be great (they work well with texts but they cannot have a different type).

I was planning on making a simple bugfix, and did not check whether some other components are necessary (like `datepicker`, `timepicker`). I'm not a master of tailwind-css, so I just tried to fix the problem that I came across. We should probably create a main `x-input.input` component, and parameterize it properly in other types of input components (like `datepicker`, `timepicker`, etc.). I will check on it.

I just now realized that I just removed the feature that can repopulate the forms in case of validation errors. I will check on that as well.

kdmnk:
Yeah these input fields can be cumbersome to use. The main reason I added those to insert the validation messages automatically as those were easy to miss out (and to avoid writing some trivial tags). I'm open to other solutions here.

When working with input fields we need to keep the followings in mind:
- validation error message are shown
- previous value from database is prefilled
- previous session data is prefilled if exists (after validation error we keep the state)

I'm still sure we don't have all these everywhere yet as there are a lot of special cases.

horcsinbalint:

PR #455, opened 2024-03-03
Small template fixes

horcsinbalint:

PR #467, opened 2024-03-11
Remove president from status cert

### LINKED ISSUE #465: Remove president from status certificate
Remove president from status certificate as requested by the secretariat

horcsinbalint:
Should fix #465


BertalanD:
Is this the only template where the changes might be problematic?

horcsinbalint:
> Is this the only template where the changes might be problematic?

In  #455 I only effectively edited this file, and deleted another. So I think it should be all.

Issue #300, opened 2023-07-19
Create resource for Documents


kdmnk:
I want to create a DocumentPolicy instead of the Gates used in AuthServiceProvider but we don't have a model yet for Documents.
It can make sense to store some of the documents in DB but we need to specify this.

PR #471, opened 2024-03-16
General assembly: excuse passives (fixes #417)

### LINKED ISSUE #417: Do not require passive students to join general assemblies
Add them to the excused users list by default.

Otherwise, they have a missing attendance after they become active again.

Trigary:


app/Models/GeneralAssemblies/GeneralAssembly.php
```diff
@@ -218,4 +220,23 @@ public static function isTemporaryPasscode(string $value): bool
...
+     */
+    protected static function boot(): void
+    {
+        parent::boot();
+
+        static::created(function ($general_assembly) {
+            // Excuse passive students. They are excused at the creation of the general assembly (as opposed to when it
+            // is opened) because the list of excused students is displayed as soon as the general assembly is created.
+            foreach (Semester::current()->usersWithStatus(SemesterStatus::PASSIVE)->get() as $user) {
+                $general_assembly->excusedUsers()->attach($user);
```
kdmnk: I believe the attach can take a collection of users in once, so you don't need the foreach.

kdmnk (changes requested):
Nice! We have some test cases for general assemblies. Can you add one to test this behaviour? Add 3 users, 1 active 2 passive, make a general assembly and check the excusedUsers if its set properly.

If you want, we could also make some improvements with the excused users here:
The table where we collect them are called "general_assembly_user" but the naming should conclude something about exclusion. Make a migration to rename it.
And another thing came into my mind is that we should add a comment/reason for the exclusion, especially if we start adding users automatically. We have some examples for pivot attributes (roles, semester status) but a simple value is enough here. You can check https://laravel.com/docs/10.x/eloquent-relationships#retrieving-intermediate-table-columns. 
Then you also need to get the value from the UI. 
Or you can open a new issue for these.

Trigary:
Thanks, I'll make the changes you mentioned.

I just saw #300 (which mentions that observers should be preferred), therefore I just pushed a commit that makes the necessary changes. Let me know if I misunderstood something and should undo the commit.

kdmnk:
No it's good, I just did not want to say it as we you only used the "created" event here. Observers are best if you handle a lot of events there but at least now we are consistent.

Trigary:
I made the requested changes :)

I made two separate database migrations, because I figured a table rename is disruptive enough to warrant its own migration file. Let me know if I should combine the migrations instead.

The way the localization is handled for the `comment` column is sub-ideal: the backend fills the database with a localized string based on the locale of the user who creates the general assembly. Instead of storing a localized string, we could add some sort of flag column that determines whether the contents of the `comment` column should be displayed as-is or if it should be used as a localization key. Let me know if the current solution is adequate or if I should do something differently.

kdmnk (approved):
LGTM. You can store the localization key and translate when you display it. It won't mess up custom values.

PR #476, opened 2024-03-17
Improve light/dark mode handling

Trigary:
Fixed bugs:

- You had to click on the light/dark mode toggle button twice to actually change from light to dark mode for the first time. (E.g. when opening the site in incognito mode.)

Other improvements:

- Exclusively local storage is used to store whether the user prefers light or dark mode: cookies are no longer used.
  - Because of this, the codebase could be simplified.
  - The downside is that the backend no longer gets notified when the user changes their preferences. But this notification is probably not needed anyway.
- The `light` class now gets added/removed from the `<body>` tag when switching between modes. Previously the `light` class would only get added if the page was originally loaded with light mode selected and wouldn't get removed even if the used changed their theme preferences.

### DIFF
--- a/app/Http/Controllers/HomeController.php
+++ b/app/Http/Controllers/HomeController.php
@@ -44,11 +44,6 @@ class HomeController extends Controller
         ]);
     }
 
-    public function colorMode($mode)
-    {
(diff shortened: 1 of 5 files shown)


kdmnk (approved):
Nice, i like it. We don't need to involve the backend.

PR #479, opened 2024-03-22
Add more PHPStorm, login and database tips

Trigary:
I think the additions are a great help for anyone starting out with the development of this project.

### DIFF
--- a/docker-dev-setup/README.md
+++ b/docker-dev-setup/README.md
@@ -48,6 +48,14 @@ docker exec -it mars_dev bash -c \
 docker compose stop
 ```
 
+### Optional setup steps
+
+These steps are not necessary for a basic setup, but might be required for some features or might be helpful in some situations.
+
+Please check out the ["Optional steps" section of the _SETUP_DEV_ENVIRONMENT.md](../_SETUP_DEV_ENVIRONMENT.md#optional-steps) file.
+You may need to change your working directory to the root (`mars`) folder, and you may need to run some commands in the `mars_dev` container.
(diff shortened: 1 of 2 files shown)


docker-dev-setup/README.md
```diff
@@ -62,6 +62,9 @@ docker exec -it mars_dev bash -c "php artisan serve --host=0.0.0.0"
 
 The app is now running at [http://localhost:8000](http://localhost:8000).
 
+You can log in as `example@eotvos.elte.hu` to gain access to a superuser account or as `collegist@eotvos.elte.hu` for a regular account.
```
kdmnk: These values can be configurable in `.env`. This is already stated at https://github.com/EotvosCollegium/mars/blob/development/_SETUP_DEV_ENVIRONMENT.md#for-everyday-use but you can put a link here pointing to this one.

docker-dev-setup/README.md
```diff
@@ -77,6 +80,8 @@ docker exec -it mars_mysql mysql --user=mars --password=password
 
 Alternatively, you can access the database directly from you host machine (via an application of your choice) at this address: `127.0.0.1:3307`
 
+You can reset the database by running `docker exec -it mars_dev bash -c "php artisan migrate:fresh --seed"`.
```
kdmnk: Isn't there a way to open docker's terminal? (at least PHPStorm should have this feature if you are using that). so that we can just forward them to the same "everyday use" part.

(also applies to the docker command for accessing the db and below to generate IDE helper files)

Trigary:
Thanks, I made the requested changes.

kdmnk (approved):

PR #481, opened 2024-03-23
User list export: apply filters (fixes #298)

### LINKED ISSUE #298: User list export - Filter by admission year, workshop, role (műhelytitkár)
Enable exporting only those users' data who were filtered (by admission year, workshop, role (műhelytitkár), etc.)

Trigary:
Summary of changes:

- The implementation behind the export button on the `/users` route has been replaced
  - The rendering has been moved from `list.blade.php` to `list_users_component.blade.php`
  - Clicking the button no longer calls the `users.export` route: the click is now handled within the livewire component
  - The same filtering logic is used within the component for the display of the users and for the exporting
- ~~To implement the filtering, a `UserFilter` utility class has been created~~
- The different `Export` classes under `UsersSheets` have been modified so that the included users can be specified
- The behavior of the `users.export` path (hopefully) hasn't changed: the same users will get included in the exports as before

Small, mostly unrelated fixes:

- The `$show_feedback` property has been deleted from `SemesterEvaluationExport`: it was unused
  - `UsersExport` also assumed that `SemesterEvaluationExport` had a `boolean` constructor parameter, while in fact, it did not have any parameters. This has also been fixed.
- `ListUsers`'s status related methods incorrectly named their parameters `status_id`, while in reality they received a status (string), not an ID


app/Utils/UserFilter.php
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * Usage:
+ * <ol>
+ * <li>Create a user query that should be further narrowed down, e.g. {@see User::canView()}</li>
+ * <li>Create a new class instance with the previously obtained query instance</li>
+ * <li>Call the desired filter methods on the instance to narrow down the query</li>
+ * <li>Get the final query by calling {@see UserFilter::getQuery()} on the instance</li>
+ * </ol>
+ */
+class UserFilter
```
kdmnk: The functionality of this class can be achieved by Laravel's scopes (see https://laravel.com/docs/10.x/eloquent#local-scopes). Rewrite it with scopes and try to create as general scopes as possible.

app/Http/Controllers/Secretariat/UserController.php
```diff
@@ -302,7 +302,7 @@ public function export()
     {
         $this->authorize('viewAny', User::class);
 
-        return Excel::download(new UsersExport(), 'uran_export.xlsx');
+        return Excel::download(new UsersExport(User::canView()), 'uran_export.xlsx');
```
kdmnk: You don't use this controller function anymore. Delete it together with it's endpoint.

app/Exports/UsersExport.php
```diff
@@ -7,20 +7,32 @@
...
+use Illuminate\Database\Eloquent\Builder;
 use Maatwebsite\Excel\Concerns\WithDefaultStyles;
 use Maatwebsite\Excel\Concerns\WithMultipleSheets;
 use PhpOffice\PhpSpreadsheet\Style\Style;
 
 class UsersExport implements WithMultipleSheets, WithDefaultStyles
 {
+    private Builder|User $includedUsersQuery;
+
+    public function __construct(Builder|User $includedUsersQuery)
```
kdmnk: why don't you pass the User collection? Using builders here makes it more complex. Order the query by name beforehand and you will have the same result.

app/Exports/UsersExport.php
```diff
@@ -7,20 +7,32 @@
...
+    {
+        $this->includedUsersQuery = $includedUsersQuery;
+    }
+
     public function sheets(): array
     {
-        $sheets = [new CollegistsExport(), new StatusesExport()];
+        $sheets = [
+            new CollegistsExport($this->includedUsersQuery->clone()),
```
kdmnk: why do you need `clone`?

resources/views/secretariat/user/list_users_component.blade.php
```diff
@@ -123,4 +123,11 @@
             </div>
         </div>
     </div>
+
+    {{-- Export --}}
+    <div class="fixed-action-btn tooltipped" data-position="left" data-tooltip="Exportálás">
```
kdmnk: Make it explicit: eg. "szűrt felhasználók adatainak exportálása"

kdmnk (changes requested):

app/Http/Controllers/Secretariat/UserController.php
Trigary: I think it's still in use: it's referenced in `evaluation_form_available_details.blade.php` and in `evaluation_form_closed.blade.php`. (Look for `<a href="{{ route('users.export') }}">letölthetőek</a>` in both files.)

app/Exports/UsersExport.php
Trigary: Different classes use the query passed to them in differing ways. By cloning, we ensure that modifications in one export class don't modify the behavior of the consequent exports.

If we were to pass a collection instead of a builder, then cloning would not be necessary. I'll use a collection.

app/Exports/UsersExport.php
Trigary: Done!

app/Utils/UserFilter.php
Trigary: Thanks!

app/Http/Controllers/Secretariat/UserController.php
kdmnk: Hm, true but it might create some confusion in the future. These are only in emails, we could change the link to point to the users index page.

Trigary: Done

app/Models/User.php
```diff
@@ -589,6 +589,65 @@ public function scopeHasToPayKKTNetregInSemester(Builder $query, int $semester_i
...
+     * The roles are specified by their IDs.
+     */
+    public function scopeHasAllRoleIds(Builder $query, array $roleIdsAll): void
+    {
+        foreach ($roleIdsAll as $roleId) {
+            $query->whereHas('roles', function (Builder $query) use ($roleId) {
+                $query->where('id', $roleId);
+            });
+        }
+    }
```
kdmnk: Please move this to App\Utils\HasRole trait. 
If you want, you could also investigate if it's possible to chain the `withRole` scope for multiple roles. Please crate some test cases if so. If not, it's fine for now. I guess it won't result in an optimal query anyways.

app/Models/User.php
```diff
@@ -589,6 +589,65 @@ public function scopeHasToPayKKTNetregInSemester(Builder $query, int $semester_i
...
+
+    /**
+     * Scope a query to only include users who are in all the specified workshops.
+     * The workshops are specified by their IDs.
+     */
+    public function scopeInAllWorkshopIds(Builder $query, array $workshopsIdsAll): void
+    {
+        foreach ($workshopsIdsAll as $workshopId) {
+            $query->whereHas('workshops', function (Builder $query) use ($workshopId) {
+                $query->where('id', $workshopId);
```
kdmnk: use whereIn and you can pass the $workshopIdsAll without the foreach

app/Models/User.php
```diff
@@ -589,6 +589,65 @@ public function scopeHasToPayKKTNetregInSemester(Builder $query, int $semester_i
...
+
+    /**
+     * Scope a query to only include users who have all the specified roles.
+     * The roles are specified by their IDs.
+     */
+    public function scopeHasAllRoleIds(Builder $query, array $roleIdsAll): void
+    {
+        foreach ($roleIdsAll as $roleId) {
+            $query->whereHas('roles', function (Builder $query) use ($roleId) {
+                $query->where('id', $roleId);
```
kdmnk: use whereIn and you can pass the $roleIdsAll without the foreach. Check the documentation to make sure that the behaviour won't be 'hasAnyRoleIds'...

kdmnk (changes requested):

app/Models/User.php
Trigary: Done! The documentation didn't help to be honest, but maybe I just didn't find the correct documentation. Testing suggests the solution indeed works.

app/Models/User.php
Trigary: Done!

app/Models/User.php
Trigary: Done!

Trigary:
The new `scopeWithAllRoles` doesn't (yet) support specifying role objects and the new tests don't use role objects. Let me know if I should change that.

kdmnk:
> The new `scopeWithAllRoles` doesn't (yet) support specifying role objects and the new tests don't use role objects. Let me know if I should change that.

Yeah I was also thinking about that but honestly I kind of want to rewrite the whole scope to support multiple roles in an optimal way and then use the scope for the hasRole function.

Whether to support object arguments or IDs is also a debatable. Objects improve readability and static typing whereas IDs are usually more optimal (if you only have the id you should be able to pass it as it is).

In short, it's fine by me, but feel free to investigate it further.

I'll take a look at this `whereHas` use case later, seems strange.

kdmnk:
FYI you can use https://laravel.com/docs/11.x/queries#debugging to see make sure the queries are good. Or also in laravel debugbar.

kdmnk (approved):
Checked the docs. LGTM, if you want to merge now.

Trigary:
I want to merge now.

PR #487, opened 2024-03-30
Version numbers updated in config

### LINKED ISSUE #429: Get version from code and not from environment
I don't 100% understand what is the rational behind having the software version numbers set manually for the given environment. (Probably it was used when some development was tested on production before it was published on the repo and setting version numbers that way was easier. While it is certainly still possible theoretically to update the code on the running server, it is rather complicated as you would need to use `sed` inside of a docker container as I have not made any graphical editors available there.) As I'm not the maintainer of the codebase but I usually deploy it to production, I find it challenging to assign proper version numbers as I don't see clearly if it should be considered a new minor release or a patch. It is also a bit weird to have the same version number for different patches. With the planned continous delivery, I don't see any way of setting minor version numbers without having it set in the codebase. I think we should either discontinue numbering minor releases (and identify each update with a commit hash), or do it in the codebase itself. I think we should also keep track of patch numbers with something described here: https://stackoverflow.com/questions/33481173/is-there-a-way-to-make-git-auto-generate-a-version-number-file-for-a-version-o

I would also like to have some area (like on the bottom left corner) where I can put information about how the software was delivered, like the environment it was prepared for, and the hash of the given commit.

kdmnk:
## Description
@horcsinbalint, if you agree, from now on you need to update the config in the PR when you merge from development to deployment
## Related Issue
Closes #429 
## Changes

- Version number updated in config.app

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,7 +3,6 @@ APP_ENV=local
 APP_KEY=
 APP_DEBUG=true
 APP_URL=http://localhost
-APP_VERSION=
 
 WIFI_DOMAIN=wifi.example.net
 WIFI_SSID=EXAMPLE_SSID
--- a/config/app.php
+++ b/config/app.php
@@ -15,13 +15,13 @@ return [
 
     'name' => env('APP_NAME', 'Urán'),
 
-    'version' => env('APP_VERSION', '3.4.1'),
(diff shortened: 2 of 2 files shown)


horcsinbalint:
Thanks, I like this. I will try it with this PR today.

PR #415, opened 2024-01-25
Trigger workflows if a PR is created, even for forks

horcsinbalint:
Should fix #414

PR #490, opened 2024-03-30
Laravel 11 upgrade

### LINKED ISSUE #488: Update to Laravel 11 and PHP 8.2

kdmnk:
## Description
Updates dependencies, laravel 11, cleanup
## Related Issue
Closes #488 
## Changes

- Updated dependencies
- Cleanup boilerplate files
- Skeleton for deployment docs
- Some extra refactoring with middlewares and gates
## TODO

- Invityr: https://github.com/EotvosCollegium/mars/pull/491


bootstrap/app.php
```diff
@@ -1,55 +1,53 @@
...
+        // api: __DIR__.'/../routes/api.php',
+        // commands: __DIR__ . '/../routes/console.php',
+        // channels: __DIR__.'/../routes/channels.php',
+        health: '/up',
+    )
+    ->withMiddleware(function (Middleware $middleware) {
+        $middleware->redirectGuestsTo('/login');
+        $middleware->trustProxies(
+            //TODO
+            //at: config('app.trust_proxies'),
```
kdmnk: @horcsinbalint I can't use config anymore to get the trusted proxies. Can you give me the list?

horcsinbalint: Yep, it is `172.16.9.200`. The code supported lists but we only have a single reverse proxy distributing the work to the different webservices. (We might change it in the future but it should not be frequent and for this purpose we use private IP addresses so we have control over them.)

.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-on: [push, pull_request]
```
horcsinbalint, line 1: Is there a reason for disabling workflows if a PR is created? I had created #415 because of contributors who do not have write permission to the repo.

kdmnk: Ah, I see. Otherwise in our case it's running twice but then I'll put it back.

app/Http/Middleware/LogRequests.php
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Handle an incoming request.
+     *
+     * @param \Closure(Request): (Response) $next
+     * @throws AuthenticationException
+     */
+    public function handle(Request $request, Closure $next): Response
+    {
+        if (Auth::check() && $request->isMethod('post')) {
+            SystemLog::debug('User #' . user()->id . ' sent request: path='
```
horcsinbalint, line 23: Does this only log in debug mode or also in production?

kdmnk: Same as before, the files just have been renamed. I think debug statements won't have an effect in production.

horcsinbalint: I checked it, a DEBUG log is created and this middleware is active in production. But it is set in multiple routes not to invoke this middleware (which is great for things like someone votes on a general assembly, authentication etc.).

routes/web.php
horcsinbalint, line 1: In line 267 `'log'` could also be changed to `LogRequests::class`

horcsinbalint: Same for line 104

kdmnk: Nice catch, thanks

Issue #503, opened 2024-04-01
Aligning text in Epistola more elegantly


viktorcsimma:
Now it’s justified, but since it is not hyphenated, it looks quite oddly on mobile (see the screenshot). It could either be left-aligned or hyphenated; the former is probably easier to solve.

![16EC6A9B-1BCD-4200-B434-96FFED9B449F](https://github.com/EotvosCollegium/mars/assets/70488334/0686541a-7ffe-4b10-8c84-7dad770c814e)

PR #508, opened 2024-04-09
Align epistola

horcsinbalint:
## Description
Fix alignment issues for epistola email
## Related Issue
#503
## Changes
- allowed hyphenation
- aligned content left for mobiles
## Additional Notes
I was unable to reproduce the issue with emails but I was able to do it on the preview

### DIFF
--- a/app/Http/Controllers/StudentsCouncil/EpistolaController.php
+++ b/app/Http/Controllers/StudentsCouncil/EpistolaController.php
@@ -5,7 +5,6 @@ namespace App\Http\Controllers\StudentsCouncil;
 use App\Http\Controllers\Controller;
 use Illuminate\Support\Facades\Validator;
 use Illuminate\Http\Request;
-use Illuminate\Support\Facades\Auth;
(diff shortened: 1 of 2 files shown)


BertalanD:
E-mail clients tend to be pretty *selective* about what CSS rules they support -- most don´t use a general-purpose browser rendering engine, see the [docs for Gmail](https://developers.google.com/gmail/design/css), for instance. 

There is also <https://caniemail.com>,  but the [data there for `hyphens`](https://www.caniemail.com/search/?s=hyphens) is pretty out-of-date.

This PR obviously isn´t going to make things any worse, but I'd double-check with real mail clients if it has any effects at all.

horcsinbalint:
@BertalanD thank you for the suggestion, I have checked it using gmail. The alignment should work with it now, I have fixed this and some other issues as well. Hyphenation is not supported on some email clients / providers (like with gmail).

viktorcsimma (approved):
Nice; thank you:)

Issue #482, opened 2024-03-24
Log the end of collegist status


viktorcsimma:
We should know when users became alumni, to be able to calculate membership counts in different years.

An option would be to simply add a timestamp to the role_users table; another would be to add a field to the educational informations.

kdmnk:
Add it to the educational_information, our roles are complex enough. Should be set automatically when we deactivate them.

kdmnk:
Also include:
- ensure that we give statuses only to active collegists

viktorcsimma:
Probably should be merged with #392.

PR #509, opened 2024-04-11
Added missing students section for general assemblies

### LINKED ISSUE #499: Add a list of missing students to general assemblies
It would be needed for the records the secretary writes. If you want, you can write it in a way only authorised people see it (so the president, the secretary and admins); but there are arguments for and against that;)

If this could be done for the next assembly on 4th April, that would be nice.

horcsinbalint:
## Description
Added missing students section
## Related Issue
#499
## Changes
- Added table for students that should attend general assemblies
- List students that are missing

### DIFF
--- a/app/Models/GeneralAssemblies/GeneralAssembly.php
+++ b/app/Models/GeneralAssemblies/GeneralAssembly.php
@@ -157,13 +157,17 @@ class GeneralAssembly extends Model
     }
 
     /**
-     * @return array The users who should have attended the general assembly but did not and were not excused.
+     * @return array The users who should have attended the general assembly but did not.
      */
-    public function missing_users(): array
-    {
+    public function missing_users(): array {
         $missing = [];
-        foreach($this->users_that_should_attend()->get() as $user) {
-            if (!$this->isAttended($user)) {
+        if($this->isOpen()){
+            $users_that_should_attend = $this->users_that_should_attend_open_assembly();
+        } else {
+            $users_that_should_attend = $this->users_that_should_attend()->get();
+        }
+        foreach($users_that_should_attend as $user) {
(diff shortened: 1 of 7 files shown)


viktorcsimma:
Maybe I miss the point, but wouldn't it be easier to connect this with #482? That way, we would know exactly who was a collegist at that time, and from the status, we would also know whether they were active in the semester.

horcsinbalint:
> Maybe I miss the point, but wouldn't it be easier to connect this with #482? That way, we would know exactly who was a collegist at that time, and from the status, we would also know whether they were active in the semester.

You are right, my implementation stores whether they were enrolled at the time which could be done easier if #482 was fixed.

kdmnk:
Just took a quick look. Yes, I don't see why storing the users in a new table is necessary. We already have everything for this in the statuses. Just make a new relation between general assemblies and semesters. Anyone who is active and not excused is supposed to be there, right?

horcsinbalint:
> Just took a quick look. Yes, I don't see why storing the users in a new table is necessary. We already have everything for this in the statuses. Just make a new relation between general assemblies and semesters. Anyone who is active and not excused is supposed to be there, right?

Yes, if alumnis are never considered active. Idk, it might be the case. I was not really sure if there could be a case where alumnis / guests have an active status as previously there was a filter in the code for non-alumni collegists. But yes, it seems to be redundant.

Is there a way for us to get the semester for a given date or we should store the semester for the GeneralAssembly object?

kdmnk:
Add a semester_id. It will fit with the new semester behaviour I'm working with.
Statuses and roles are not connected but without human intervention we should not have an active alumni. Worst case, someone will find them in this list and update them so it's not a big deal.

horcsinbalint:

PR #512, opened 2024-04-15
Implement KKT role

### LINKED ISSUE #484: Add a role for those who administer kkt payments
Actually, there already exists a role for this: cultural committee members can administer kkt:D (see [the code](https://github.com/EotvosCollegium/mars/blob/b35d6717045526ec26b9caf9ffda473551463a8e/app/Policies/CheckoutPolicy.php#L50); this has historical reasons btw). It would be great to have a separate role (e.g. "kkt handler" or something like that) which the economic vice president (and admins) can give and take away.

imol-ai:
## Description
Adds the KKT role, which shall be assigned to the persons handling the payment.
## Related Issue
Closes #484 
## Changes
- Adds a kkt-handler role
- When having this, role, one can register payments in /economic_committee
- The economic vice president can assign this role

### DIFF
--- a/app/Models/Role.php
+++ b/app/Models/Role.php
@@ -57,6 +57,7 @@ class Role extends Model
     public const CULTURAL_LEADER = 'cultural-leader';
     public const CULTURAL_REFERENT = 'cultural-referent';
     public const CULTURAL_MEMBER = 'cultural-member';
+    public const KKT_HANDLER = 'kkt-handler';
     public const COMMUNITY_LEADER = 'community-leader';
     public const COMMUNITY_REFERENT = 'community-referent';
     public const COMMUNITY_MEMBER = 'community-member';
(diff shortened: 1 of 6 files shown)


app/Models/Role.php
```diff
@@ -96,7 +97,8 @@ class Role extends Model
         self::STUDENT_COUNCIL_SECRETARY,
         self::STUDENT_COUNCIL,
         self::BOARD_OF_TRUSTEES_MEMBER,
-        self::ETHICS_COMMISSIONER
+        self::ETHICS_COMMISSIONER,
+        self::KKT_HANDLER,
```
viktorcsimma: Maybe this could go under `COMMITTEE_MEMBERS`; they were originally members of the Economic Committee, and they belong under the economic vice president, rather than being independent office holders. This way, it would be under the "Students Council" dropdown, which might be a more logical choice. But maybe this is only my preference:)

viktorcsimma (changes requested):
Nice:) If you have the time, could you move it under the Students Council role? See my comment; sorry for not being concrete enough.

imol-ai:
Thanks for the comments, fixed.

viktorcsimma (approved):
Sorry for reviewing late; it's nice:) Thank you! Merge as you wish;)

PR #518, opened 2024-04-20
add database to connect command mysql

Rreebb:
Use automatically the mars database when opening mysql.


viktorcsimma (approved):
Nice catch; thank you:)

PR #520, opened 2024-04-24
Adding the secretariat access to general assemblies

viktorcsimma:
The secretariat has asked for access to general assembly results (they are already going to see the records, so this is only for convenience). Could we deploy this as soon as possible?


kdmnk (approved):

viktorcsimma:
Thank you:)

PR #526, opened 2024-05-09
Correcting an information text about averages in evaluation form

viktorcsimma:
Dávid Szaniszló has told me about this: the original text was probably not what we meant.

This is just changing a text in a Blade template; hopefully, we can deploy it before we reopen the form.

## Summary by CodeRabbit

- **Documentation**
	- Updated instructional message in the evaluation form to clarify grade input timing related to deadlines.

### DIFF
--- a/resources/views/secretariat/evaluation-form/avg.blade.php
+++ b/resources/views/secretariat/evaluation-form/avg.blade.php
@@ -1,7 +1,7 @@
 <form method="POST" action="">
     @csrf
     <blockquote>
-        Ha még nem vagy lezárva minden tárgyból, ne írd be az átlagodat! (Ha lezárnak a beküldési határidő előtt, akkor be tudod majd írni akkor is.)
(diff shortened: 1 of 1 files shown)


resources/views/secretariat/evaluation-form/avg.blade.php
```diff
@@ -1,7 +1,7 @@
 <form method="POST" action="">
     @csrf
     <blockquote>
-        Ha még nem vagy lezárva minden tárgyból, ne írd be az átlagodat! (Ha lezárnak a beküldési határidő előtt, akkor be tudod majd írni akkor is.)
```
kdmnk, line 4: I originally meant that they can edit the form as many times as they want before the deadline.
The data is connected to the form, so they won't be able to edit this after the deadline.

kdmnk (changes requested):

resources/views/secretariat/evaluation-form/avg.blade.php
viktorcsimma: Oh, I see... what about this? _"A beküldési határidőig bármikor beírhatod majd."_

viktorcsimma:
Finally addressed in #552.

PR #528, opened 2024-05-11
Fix default value for select

kdmnk:
## Description
The current value was overwritten if the default value was given for the select component.

## Summary by CodeRabbit

- **New Features**
	- Enhanced the select input component to remember previously entered values after form submission.


kdmnk:
I'm merging this without a review in a reasonable time. It's only an easy fix.

PR #534, opened 2024-05-31
filter last 2 general assemblies for new students

### LINKED ISSUE #416: Filter last 2 general assemblies for new students
Freshmen are not expected to join the general assemblies before they joined.

csandras05:
Filter the last 2 general assemblies for freshmen as they are not expected to join the general assemblies before they joined.

If the date of acceptance (September 1, the year of acceptance) is later than the second last general assembly, then the user automatically passed the requirements.

Closes #416 

## Summary by CodeRabbit

- **New Features**
  - Enhanced the requirements check for general assemblies to include logic for new students based on their acceptance date and educational information.

- **Tests**
  - Added a new test to verify that new students meet the requirements for general assemblies.

### DIFF
--- a/app/Models/GeneralAssemblies/GeneralAssembly.php
+++ b/app/Models/GeneralAssemblies/GeneralAssembly.php
@@ -3,6 +3,7 @@
 namespace App\Models\GeneralAssemblies;
 
 use App\Enums\PresenceType;
+use App\Models\EducationalInformation;
 use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Model;
@@ -147,7 +148,12 @@ class GeneralAssembly extends Model
      */
     public static function requirementsPassed(User $user): bool
     {
-        $lastAssemblies = GeneralAssembly::all()->sortByDesc('closed_at')->take(2);
+        $year_of_acceptance = $user->educationalInformation->year_of_acceptance;
+        $acceptance_date = now()->setYear($year_of_acceptance)->setMonth(9)->setDay(1);
+        $lastAssemblies = GeneralAssembly::orderBy('closed_at', 'desc')->take(2);
+        if ($acceptance_date > $lastAssemblies->get()[1]->closed_at) { // If the user was accepted after the second last assembly
+            return true;
+        }
         foreach ($lastAssemblies as $assembly) {
             if ($assembly->isAttended($user)) {
                 return true;
--- a/app/Models/GeneralAssemblies/GeneralAssembly.php
+++ b/app/Models/GeneralAssemblies/GeneralAssembly.php
@@ -10,6 +10,7 @@ use Illuminate\Database\Eloquent\Model;
 
 use App\Models\GeneralAssemblies\Question;
 use App\Models\User;
+use Carbon\Carbon;
 use Illuminate\Contracts\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Support\Collection;
@@ -149,10 +150,12 @@ class GeneralAssembly extends Model
     public static function requirementsPassed(User $user): bool
     {
         $year_of_acceptance = $user->educationalInformation->year_of_acceptance;
-        $acceptance_date = now()->setYear($year_of_acceptance)->setMonth(9)->setDay(1);
-        $lastAssemblies = GeneralAssembly::orderBy('closed_at', 'desc')->take(2);
-        if ($lastAssemblies->count() >= 2 && $acceptance_date > $lastAssemblies->get()[1]->closed_at) {
-            // If the user was accepted after the second last assembly
+        $acceptance_date = Carbon::createFromDate($year_of_acceptance, 9, 1);
+
+        $lastAssemblies = GeneralAssembly::orderBy('closed_at', 'desc')->take(2)->get();
+        $secondLastAssembly = $lastAssemblies->count() >= 2 ? $lastAssemblies[1] : null;
+
+        if ($secondLastAssembly && $secondLastAssembly->closed_at < $acceptance_date) {
             return true;
         }
         foreach ($lastAssemblies as $assembly) {
--- a/app/Models/GeneralAssemblies/GeneralAssembly.php
+++ b/app/Models/GeneralAssemblies/GeneralAssembly.php
@@ -151,7 +151,8 @@ class GeneralAssembly extends Model
         $year_of_acceptance = $user->educationalInformation->year_of_acceptance;
         $acceptance_date = now()->setYear($year_of_acceptance)->setMonth(9)->setDay(1);
         $lastAssemblies = GeneralAssembly::orderBy('closed_at', 'desc')->take(2);
-        if ($acceptance_date > $lastAssemblies->get()[1]->closed_at) { // If the user was accepted after the second last assembly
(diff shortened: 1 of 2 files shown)


viktorcsimma:
The idea looks good to me... can you check out the failing tests? They complain about a unit test and a documentation comment.

app/Models/GeneralAssemblies/GeneralAssembly.php
```diff
@@ -147,7 +148,13 @@ public function isAttended(User $user): bool
      */
     public static function requirementsPassed(User $user): bool
     {
-        $lastAssemblies = GeneralAssembly::all()->sortByDesc('closed_at')->take(2);
+        $year_of_acceptance = $user->educationalInformation->year_of_acceptance;
+        $acceptance_date = now()->setYear($year_of_acceptance)->setMonth(9)->setDay(1);
+        $lastAssemblies = GeneralAssembly::orderBy('closed_at', 'desc')->take(2);
```
viktorcsimma: This returns an empty collection for me... could you check out `GeneralAssembly::all()->sortByDesc('closed_at')->take(2)`? I found it in SemesterEvaluationController, it seems to work.

viktorcsimma (changes requested):
Something did not work for me; please see my comments.

Also, could you open the evaluation form? For me, it worked by temporarily telling isEvaluationAvailable() in SemesterEvaluationController to always return true. That way, you can try it by logging in with [collegist@eotvos.elte.hu](mailto:collegist@eotvos.elte.hu) and then opening .../secretariat/evaluation in the browser.

tests/Unit/GeneralAssemblyTest.php
```diff
@@ -166,4 +167,25 @@ public function test_passive_users_get_excused_automatically(): void
...
+        $generalAssembly = GeneralAssembly::factory()->create();
+        $generalAssembly2 = GeneralAssembly::factory()->create();
+
+        $generalAssembly->update(['closed_at' => now()->setYear(now()->year)->setMonth(2)->setDay(15)]);
+        $generalAssembly2->update(['closed_at' => now()->setYear(now()->year)->setMonth(9)->setDay(15)]);
+
+        $generalAssembly->presenceChecks()->create();
+        $generalAssembly2->presenceChecks()->create();
+
+        $this->assertTrue(GeneralAssembly::requirementsPassed($user));
```
viktorcsimma, line 186: Also, it would be nice if you could add a case where the user does _not_ pass the requirements; e.g. by adding one more assembly in September.

app/Models/GeneralAssemblies/GeneralAssembly.php
csandras05: I changed `sortByDesc` to `orderBy` because using the former when I tried to access `$lastAssemblies[1]` (the second last assembly), I found that `$lastAssemblies$` wasn't sorted. (Although the elements seemed to come in order when I tried to iterate it with foreach).

tests/Unit/GeneralAssemblyTest.php
csandras05: Done

app/Models/GeneralAssemblies/GeneralAssembly.php
kdmnk: Does take execute the query? Just because you have the ->get() call afterwards. Move the ->get to the $lastAssemblies initialiser, in any case.

kdmnk: Use Caron::createFromDate($year_of_acceptance, 9, 1)

app/Models/GeneralAssemblies/GeneralAssembly.php
```diff
@@ -147,7 +148,13 @@ public function isAttended(User $user): bool
...
+        if ($lastAssemblies->count() >= 2 && $acceptance_date > $lastAssemblies->get()[1]->closed_at) {
```
kdmnk: move second last assembly to a new var for better readability, then you don't even need the comment

tests/Unit/GeneralAssemblyTest.php
```diff
@@ -166,4 +167,28 @@ public function test_passive_users_get_excused_automatically(): void
...
     }
+
+    /**
+     * @return void
+     */
+    public function test_new_students_pass_requirements(): void
+    {
+        $user = User::factory()->create(['verified' => true]);
+        $user->educationalInformation()->save(EducationalInformation::factory()->make());
+        $user->educationalInformation->update(['year_of_acceptance' => now()->year]);
```
kdmnk:
```suggestion
        EducationalInformation::factory()->create(['user_id' => $user->id, 'year_of_acceptance' => now()->year]);
```

tests/Unit/GeneralAssemblyTest.php
```diff
@@ -166,4 +167,28 @@ public function test_passive_users_get_excused_automatically(): void
...
+
+        $generalAssembly = GeneralAssembly::factory()->create();
```
kdmnk:
```suggestion
        $generalAssembly = GeneralAssembly::factory()->create(['closed_at' => Carbon::createFromDate(now()->year, 2, 15);
```
and delete the update. Do the same for the second assembly.

tests/Unit/GeneralAssemblyTest.php
```diff
@@ -166,4 +167,28 @@ public function test_passive_users_get_excused_automatically(): void
...
+        $generalAssembly2 = GeneralAssembly::factory()->create();
+
+        $generalAssembly->update(['closed_at' => now()->setYear(now()->year)->setMonth(2)->setDay(15)]);
+        $generalAssembly2->update(['closed_at' => now()->setYear(now()->year)->setMonth(9)->setDay(15)]);
+
+        $generalAssembly->presenceChecks()->create();
```
kdmnk, line 183: Technically, the user should pass even without presenceChecks, shouldn't? Or if you need them, please add a comment why.

kdmnk (changes requested):

kdmnk:
I left some minor comments mostly regarding readability. Good job!

app/Models/GeneralAssemblies/GeneralAssembly.php
kdmnk: Ps. Use 'orderBy' for query builders, 'sortBy' for collections (after values are queried).

kdmnk (approved):

tests/Unit/GeneralAssemblyTest.php
csandras05: Before the modification to `GeneralAssembly.php/requirementsPassed`, the user would have failed the assert after the presenceChecks, because they didn't take part in the last two general assemblies. However, note that the second last general assembly in the test (02.15) is before the acceptance date of the user (09.01), so the user shouldn't be expected to have taken part in that GA, thus they should pass the requirements.

csandras05: Fun fact: if the number of presence checks during a GA is 2, the users are expected to take part in both of them. However, when the number of presence checks is 3, the users are expected to take part in only 1, cf.
https://github.com/EotvosCollegium/mars/blob/cedc643006735db86b395d522f91866f97687d23/app/Models/GeneralAssemblies/GeneralAssembly.php#L199-L203
I don't know what would be the expected behavior here, should I open an issue regarding this?

sszajbely: I think you can simply use `$presenceCheckCount <= 2 ? $presenceCheckCount : max($presenceCheckCount - 2, 2)`.

kdmnk: Just make sure to update any description for the users, if we state the calculation method anywhere.

PR #532, opened 2024-05-16
Introducing PeriodicEvents instead of EventTriggers

kdmnk:
See https://github.com/EotvosCollegium/mars/issues/421#issuecomment-2035019754 for the description behind the idea.
The HasPeriodicEvent trait is going to be used by controllers (eg. ApplicationController)
Other PRs will come to update the existing logic (applications, semester evaluation, etc) one by one. 
I just seperated it to multiple PRs, which also means that some test cases are commented out temporarily.
I'll post these in some days in case it's hard to understand without the use cases (or you can check here: https://github.com/EotvosCollegium/mars/pull/525).
Please let me know if you have questions so I can improve the readablity

## Summary by CodeRabbit

- **New Features**
  - Introduced periodic event handling, including start, reminder, and end actions.
  - Added new email notifications for evaluation form availability, reminders, and closure.
  - Added functionality to manage voting periods and application deadlines.

- **Bug Fixes**
  - Updated validation keys and translations in English and Hungarian language files.

- **Tests**
  - Enhanced test coverage for periodic events and status updates.
  - Added new test methods for `PeriodicEvent` and related functionalities.

- **Chores**
  - Updated `.gitignore` to include new directories.
  - Adjusted environment variable configurations in `.env.example`.

- **Documentation**
  - Updated documentation to reflect new methods and functionalities for periodic events and evaluations.

PR #542, opened 2024-06-06
Anonymous questions

### LINKED ISSUE #530: Add new anonymous questions to the evaluation form
We've got a request to add some [new questions](https://docs.google.com/forms/d/1VxKF_DznSfeyq9TND2q8QhOMt8IqLxzGn7B8MDMZh6M/edit). These, however, should be anonymous and we should only store the _year of admission_ and the _status_ (resident or extern) with the answers.

The easiest solution would be to use the same mechanism as for the feedback question now (i.e. sending an email). A bit more advanced way is to add a single table (let's say, `anonymous_answers`), with a column for the year, the status and each question. A more difficult, but also more flexible approach would be to also create a table for the questions themselves, which could be editable without changing the code.

@kdmnk what do you think?

viktorcsimma:
## Description
Adds anonymous questions to evaluation forms, which can be managed by the Students' Council
and whose results can be exported into an Excel spreadsheet.
## Related Issue
Closes #530.
## Changes
- The idea was reusing the Question class from general assemblies with a polymorphic parent relationship (thanks @kdmnk). Answers of an individual are stored under so-called "answer sheets", connected to question options and long answers, and also containing some information (for now, only the year of acceptance).
- On a panel under _Students' Council_, authorized people (i.e. the president, the scientific vice president and the admins) can list the semesters, create new questions for a semester and export results.
- Under the evaluation form, anonymous questions are the last point before setting the status. Here, the user can select their answers (there can also be ones with long, written answers) and save them. They cannot answer more than once, because the fact that they have voted does get recorded (similarly as in the voting system).
## Note
It would be nice if we could merge it until Sunday; the form should be opened next week. Thanks in advance:) I will check out #532 until then.


horcsinbalint:
Thank you, I just quickly ran through the code, and the code quality seems to be quite nice.
Before checking the details of it, I would like to ask some questions to know to what is the exact goal that we want to achieve:

- Have we got permission from the director that we effectively fire students if they don't answer the questions from the Student Council?
-  What does anonymous mean in this context? Does it mean that the Student Council cannot reconstruct who have submitted the answers for each question? Or does it mean that it is impossible to reconstruct this information from the database's nominal state after data collection has been stopped? (In case of Mr. & Miss, it is possible to get the votes using the database, but in the case of General Assemblies, it is impossible to pair the votes with the votee after the voting AFAIK.) _I call the information which can be acquired with some `SELECT` queries from the tables that store the data the nominal state of the database._

viktorcsimma:
@horcsinbalint thanks for checking it out so quickly:)
- The first one is probably not going to happen; I think it's going to be fine this way (every question will include an option like "I don't want to answer").
- For the second one: the goal is that no one be able to connect these answers to people (including admins). I don't know whether the logging feature which has caused problems at voting means an obstacle here. Another point is that if there is a year from which only one student has remained (there is one now), they obviously can be tracked.

horcsinbalint:
@viktorcsimma
> The first one is probably not going to happen; I think it's going to be fine this way (every question will include an option like "I don't want to answer").

I have concerns with making it mandatory to answer anything on these (even if there is an option of "I don't want to answer").

I think I might have spotted some bugs in the code. But as I have major concerns with the planned functionality proposed in this PR, I am unable to process the PR further. Unfortunately, the automatic deployment is not working and I will have to manually sign these changes off. Is it okay with you if I consult with the Ethics Commissioners to put my worries to rest? If there are no issues with the proposed functionality, it should not be a problem. But if they share my concerns, it is much better to resolve the problems before I deploy these changes.

viktorcsimma:
Huh; the problem is that it would be needed ASAP… what if I took personal responsibility for it? Or if we looked for someone else to approve it? @kdmnk?

horcsinbalint:
Okay, I can parallelize processing this issue and discussing my concerns with the Ethics Commissioners if it is okay with you.

routes/web.php
```diff
@@ -266,4 +267,15 @@
...
     Route::post('/general_assemblies/{general_assembly}/presence_checks/{presence_check}/close', [GeneralAssemblyPresenceCheckController::class, 'closePresenceCheck'])->name('general_assemblies.presence_checks.close');
     Route::post('/general_assemblies/{general_assembly}/presence_checks/{presence_check}/sign_presence', [GeneralAssemblyPresenceCheckController::class, 'signPresence'])->name('general_assemblies.presence_checks.presence.store')->withoutMiddleware(LogRequests::class);
+
+    /** anonymous questions */
+    Route::prefix('/anonymous_questions')->name('anonymous_questions.')->group(function () {
+        Route::get('/', [AnonymousQuestionController::class, 'indexSemesters'])->name('index_semesters');
+        Route::get('/{semester}/questions/', [AnonymousQuestionController::class, 'index'])->name('index');
+        Route::get('/{semester}/questions/create', [AnonymousQuestionController::class, 'create'])->name('create');
+        Route::post('/{semester}/questions', [AnonymousQuestionController::class, 'store'])->name('store');
```
horcsinbalint, line 293: It should not have the logging middleware

viktorcsimma:
OK, thank you:)

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
```diff
@@ -0,0 +1,70 @@
...
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('answer_sheets', function (Blueprint $table) {
+            $table->id();
+            $table->unsignedSmallInteger('semester_id');
+            // $table->boolean('is_resident'); // but the BBs... let this be a question
+            $table->unsignedSmallInteger('year_of_acceptance');
+            // no timestamps (for better anonymity)
```
horcsinbalint: The order in which answers are collected can be used to break the anonymity

resources/lang/hu/anonymous_questions.php
```diff
@@ -0,0 +1,17 @@
...
+return [
+    'all_questions_filled' => 'Már kitöltötted ezeket a kérdéseket; köszönjük!',
+    'anonymous_questions' => 'Anonim visszajelzések',
+    'create_question' => 'Kérdés hozzáadása',
+    'export' => 'Exportálás',
+    'has_long_answers' => 'kifejtős',
+    'information_text' => "Ezek a visszajelzések a Választmánynak segítenek,
+                            hogy jobb munkát végezzen.<br />
+                            A válaszaid névtelenek
+                            (csak a tényt tároljuk, hogy válaszoltál),
```
horcsinbalint: If we store their admission year, then it should also reflect this

resources/lang/en/anonymous_questions.php
```diff
@@ -0,0 +1,17 @@
...
+return [
+    'all_questions_filled' => 'You have filled all these questions; thank you!',
+    'anonymous_questions' => 'Anonymous questions',
+    'create_question' => 'Create question',
+    'export' => 'Export',
+    'has_long_answers' => 'expects long, written answers',
+    'information_text' => "The following questions are going to be used to assess
+                            how the Students' Council could improve their work.<br />
+                            Your answers are anonymous
+                            (only the fact that you have answered is recorded),
```
horcsinbalint: If we store their admission year, then it should also reflect this

horcsinbalint:
Wait, is it mandatory to answer anonymous questions? If not, then I have no big concerns

horcsinbalint (dismissed):
I have found some ways of breaking the anonymity. I have not checked the PR in details, so others should also check the PR

resources/lang/hu/anonymous_questions.php
viktorcsimma: Done:)

resources/lang/en/anonymous_questions.php
viktorcsimma: Done:)

routes/web.php
viktorcsimma: If I'm right, that must be done by adding `->withoutMiddleware([LogRequests::class])`. I've added that now; thanks:)

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
viktorcsimma: Huh; I see... could this somehow be solved quickly? I have no good ideas for that; besides adding a random id for each answer sheet, long answer and answer sheet-option connection... and even then, it is probably not cryptographically secure:D

routes/web.php
viktorcsimma: (At least for the ones which are related to filling the form; I think the others can retain logging.)

horcsinbalint: Yeah, I also wanted to only mark that

viktorcsimma:
> Wait, is it mandatory to answer anonymous questions? If not, then I have no big concerns

Now, it's not technically mandatory to fill them; my personal opinion is that we need not explicitly tell fillers about it, as they otherwise won't fill:D But could you talk to the Ethics Commissioners since then?

> I have found some ways of breaking the anonymity. I have not checked the PR in details, so others should also check the PR

Is there anyone here who could check this out ASAP? (I know, I was slow in responding to requests; sorry)

horcsinbalint:
> Now, it's not technically mandatory to fill them; my personal opinion is that we need not explicitly tell fillers about it, as they otherwise won't fill:D But could you talk to the Ethics Commissioners since then?

I will respond to the other thread soon which probably needs to be addressed.
As it has been discussed privately, I did not escalate this issue to the Ethics Commissioners, as I don't think that I would violate ethical norms by deploying this.

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
horcsinbalint: Well, this has been a question for me for quite some time. And that is why I proposed to simply use an embedded Google Forms sheet.

The order in which records are created are probably stored regardless of the identifier so randomizing it is probably not enough.

In theory, it is possible to anonymously collect the answers. These are **not** a recommendations:

- we could create a record for each user when the question is created where it is indicated whether they have already voted. The record can be updated when they vote. The order on which these updates are created is probably not possible to retrieve afterwards if binary logging is disabled. (If binary logging is enabled, than it is trivial to restore.) Probably it would still be possible using forensic tools but I intentionally wrote about the nominal state of the schemas.
- currently, we add a new record when a user submits an answer. After this, we could create a random permutation of `user_id`s. It has the same limitations as I the previous solution. And it also increases processing time, and one would need to be sure to make this procedure atomic, consistent and isolated from other requests which are not trivial (not just throwing the code into a transaction as it can cause dead locks).

As always, a miscreant system administrator can simply wiretap the HTTP requests to access the information.

I would probably not implement any of my proposed solutions. They seem to be quite cumbersome and not easy to implement, and I just wanted to show that it **is** possible to get to a point where it is extremely difficult to restore who answered what. And also it only limits people with who have direct control over the server which is quite a small group.

This issue is easier to solve when data is aggregated during data collection.

Some resolutions:
- we simply accept that system administrators do see who answered what
- we do not store the fact whether a user has submitted an answer

viktorcsimma:
OK; thank you:)

app/Exports/UsersSheets/AnonymousQuestionsExport.php
```diff
@@ -0,0 +1,78 @@
...
+    public function map($answerSheet): array
+    {
+        $row = [
+            $answerSheet->id,
+            $answerSheet->semester->tag,
+            $answerSheet->year_of_acceptance
+        ];
+        foreach ($this->semester->questions()->orderBy('id')->get() as $question) {
+            if ($question->has_long_answers) {
+                $row[] = $answerSheet->longAnswers()
```
kdmnk: If this is the only place where we display the responses, it's fine. If there's a possibility we might need it elsewhere later, add a function for the model to retrieve its string representation.

app/Http/Controllers/Secretariat/AnonymousQuestionController.php
```diff
@@ -0,0 +1,202 @@
...
+
+    /**
+     * Returns a page with the options (and results, if authorized) of a question.
+     */
+    public function show(Semester $semester, Question $question)
+    {
+        $this->authorize('administer', AnswerSheet::class);
+        // check whether it really belongs here
+        // and throw a 404 if not
+        if ($semester != $question->parent) {
```
kdmnk: Check https://laravel.com/docs/10.x/routing#implicit-model-binding-scoping

app/Http/Controllers/Secretariat/AnonymousQuestionController.php
```diff
@@ -0,0 +1,202 @@
...
+        }
+        return view('anonymous_questions.show', [
+            "question" => $question
+        ]);
+    }
+
+    /**
+     * Returns validation rules for the semester's anonymous questions.
+     */
+    private static function answerValidationRules(Semester $semester): array
```
kdmnk: I don't like this. Separate messing with arrays/keys and generating the Rules into different functions.

app/Http/Controllers/Secretariat/AnonymousQuestionController.php
```diff
@@ -0,0 +1,202 @@
...
+    {
+        $rules = [];
+        foreach ($semester->questionsNotAnsweredBy(user()) as $question) {
+            // It does not really work well with ids as keys;
```
kdmnk: Make an attribute for the questions to get the key used for the requests. Use that in the frontend too. You don't need the comment then.

app/Http/Controllers/Secretariat/AnonymousQuestionController.php
```diff
@@ -0,0 +1,202 @@
...
+    public function storeAnswerSheet(Request $request, Semester $semester)
+    {
+        $this->authorize('is-collegist');
+
+        $validatedData = $request->validate(self::answerValidationRules($semester));
+
+        // Since answer sheets are anonymous,
+        // we cannot append new answers to the previous sheet (if any);
+        // we have to create a new one.
+        $answerSheet = AnswerSheet::createForUser(user(), $semester);
```
kdmnk: why do you pass user() for every function? Use user() within.

app/Http/Controllers/StudentsCouncil/GeneralAssemblyQuestionController.php
```diff
@@ -72,10 +74,14 @@ public function store(Request $request, GeneralAssembly $generalAssembly)
     /**
      * Returns a page with the options (and results, if authorized) of a question.
      */
-    public function show(GeneralAssembly $generalAssembly, $question)
+    public function show(GeneralAssembly $generalAssembly, Question $question)
     {
         $this->authorize('viewAny', GeneralAssembly::class);
-        $question = $generalAssembly->questions()->findOrFail($question);
+        // check whether it belongs here
+        // and throw a 404 if not
```
kdmnk: same as before

app/Http/Controllers/StudentsCouncil/GeneralAssemblyQuestionController.php
```diff
@@ -84,16 +90,20 @@ public function show(GeneralAssembly $generalAssembly, $question)
...
      * Opens a question.
      */
-    public function openQuestion(GeneralAssembly $generalAssembly, $question)
+    public function openQuestion(GeneralAssembly $generalAssembly, Question $question)
     {
         $this->authorize('administer', GeneralAssembly::class);
         $this->authorize('administer', GeneralAssembly::class);
-        $question = $generalAssembly->questions()->findOrFail($question);
+        // check whether it really belongs here
+        // and throw a 404 if not
```
kdmnk: same as before

app/Http/Controllers/StudentsCouncil/GeneralAssemblyQuestionController.php
```diff
@@ -102,12 +112,16 @@ public function openQuestion(GeneralAssembly $generalAssembly, $question)
     /**
      * Closes a question.
      */
-    public function closeQuestion(GeneralAssembly $generalAssembly, $question)
+    public function closeQuestion(GeneralAssembly $generalAssembly, Question $question)
     {
         $this->authorize('administer', GeneralAssembly::class);
-        $question = $generalAssembly->questions()->findOrFail($question);
+        // check whether it really belongs here
+        // and throw a 404 if not
```
kdmnk: same

app/Http/Controllers/StudentsCouncil/GeneralAssemblyQuestionController.php
```diff
@@ -116,9 +130,13 @@ public function closeQuestion(GeneralAssembly $generalAssembly, $question)
     /**
      * Saves a vote.
      */
-    public function saveVote(Request $request, GeneralAssembly $generalAssembly, $question)
+    public function saveVote(Request $request, GeneralAssembly $generalAssembly, Question $question)
     {
-        $question = $generalAssembly->questions()->findOrFail($question);
+        // check whether it really belongs here
+        // and throw a 404 if not
```
kdmnk: same

app/Models/AnonymousQuestions/AnswerSheet.php
```diff
@@ -0,0 +1,75 @@
...
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use Illuminate\Database\Eloquent\Model;
+
+use App\Models\User;
+use App\Models\Semester;
+use App\Models\GeneralAssemblies\QuestionOption;
+use App\Models\AnonymousQuestions\LongAnswer;
+
+/**
+ * Groups a collegist's answers to anonymous questions
```
kdmnk, line 17: I don't understand. What's the connection here to the questions? How many questions does this include?

app/Models/GeneralAssemblies/Question.php
```diff
@@ -47,16 +52,27 @@ class Question extends Model
...
+
+    /**
+     * Whether the parent is a general assembly
+     * or a semester (evaluation form).
      */
-    public function generalAssembly(): BelongsTo
+    public function isForAssembly(): bool
     {
-        return $this->belongsTo(GeneralAssembly::class);
+        return 'App\\Models\\GeneralAssemblies\\GeneralAssembly' == $this->parent_type;
```
kdmnk: doesn't GeneralAssembly::class work here?

app/Models/GeneralAssemblies/Question.php
```diff
@@ -7,15 +7,20 @@
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Model;
 
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
```
kdmnk: move this outside of the scope of GeneralAssemblies.

app/Models/GeneralAssemblies/Question.php
```diff
@@ -47,16 +52,27 @@ class Question extends Model
...
+    {
+        return $this->morphTo();
+    }
+
+    /**
+     * Whether the parent is a general assembly
+     * or a semester (evaluation form).
      */
-    public function generalAssembly(): BelongsTo
+    public function isForAssembly(): bool
```
kdmnk: The Question should be as simple as general as possible. It should not care about it's parent in this kind of relationship. Refactor so that we can delete this function.

app/Models/GeneralAssemblies/Question.php
```diff
@@ -7,15 +7,20 @@
...
 use App\Models\GeneralAssemblies\QuestionUser;
+use App\Models\AnonymousQuestions\AnswerSheet;
+use App\Models\AnonymousQuestions\LongAnswer;
 use App\Models\User;
-use Illuminate\Database\Eloquent\Relations\BelongsToMany;
-use Throwable;
+use App\Models\Semester;
 
 /**
  * App\Models\GeneralAssemblies\Question
```
kdmnk: Add comment what is this and what type of questions are these (multiple choice etc)

app/Models/GeneralAssemblies/Question.php
kdmnk: For the isOpened/Active etc states, simplify these and add further checks or ignore them in GeneralAssemblies or the Answersheet/SemesterEvaluation

app/Models/GeneralAssemblies/Question.php
```diff
@@ -106,7 +136,7 @@ public function isClosed(): bool
      */
     public function open(): void
     {
-        if (!$this->generalAssembly->isOpen()) {
+        if (!$this->parent->isOpen()) {
```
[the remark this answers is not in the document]
kdmnk: add check in the parent

app/Models/GeneralAssemblies/Question.php
```diff
@@ -179,4 +217,68 @@ public function vote(User $user, array $options): void
...
+                QuestionUser::create([
+                    'question_id' => $this->id,
+                    'user_id' => $user->id,
+                ]);
+                foreach ($options as $option) {
+                    if ($option->question_id != $this->id) {
+                        throw new Exception("Received an option which does not belong to the question");
+                    }
+                    $option->increment('votes');
+                    DB::table('answer_sheet_question_option')->insert([
```
kdmnk: what's this for? we don't have in the other function. Or is it in an observer?

app/Models/GeneralAssemblies/Question.php
```diff
@@ -179,4 +217,68 @@ public function vote(User $user, array $options): void
...
+     * Creates a long answer for the question
+     * and records that the user has answered the question.
+     * Throws if the question does not accept long answers.
+     */
+    public function giveLongAnswer(User $user, AnswerSheet $answerSheet, string $text): LongAnswer
+    {
+        if (!$this->has_long_answers) {
+            throw new \Exception('this question does not accept long answers');
+        } else {
+            QuestionUser::create([
```
kdmnk: DB::transaction has been left out

app/Models/GeneralAssemblies/Question.php
```diff
@@ -148,14 +181,19 @@ public function hasVoted(User $user): bool
...
-     * Votes for a list of given options in the name of the user.
+     * Votes for an option or a list of given options in the name of the user.
      * @param User $user
-     * @param array $options QuestionOption array
+     * @param QuestionOption|array $options
      * @throws Exception if an option does not belong to the question or if too many options are selected.
      * @throws Throwable
      */
-    public function vote(User $user, array $options): void
+    public function vote(User $user, QuestionOption|array $options): void
```
kdmnk: I prefer to have one "storeAnswers" function which can handle both GeneralAssemblies/AnonymousQuestions and text/options.

app/Models/Semester.php
```diff
@@ -152,18 +159,26 @@ public function getEndDate(): Carbon
         return Carbon::createFromDate($year, $month, 1)->endOfWeek();
     }
 
+    /**
+     * Whether the semester is in the past.
+     */
+    public function isClosed(): bool
```
kdmnk, line 165: With the new periodic event handling, we should not rely on exact dates of our semester representation

app/Policies/AnswerSheetPolicy.php
```diff
@@ -0,0 +1,23 @@
...
+class AnswerSheetPolicy
+{
+    /**
+     * Whether the user can create questions,
+     * access results etc.
+     */
+    public function administer(User $user): bool
+    {
+        return $user->isAdmin()
+          || User::studentCouncilLeaders()->contains($user);
```
kdmnk: use $user->hasRole(...)

app/Http/Controllers/Secretariat/AnonymousQuestionController.php
```diff
@@ -0,0 +1,202 @@
...
+        ]);
+    }
+
+    /**
+     * Returns validation rules for the semester's anonymous questions.
+     */
+    private static function answerValidationRules(Semester $semester): array
+    {
+        $rules = [];
+        foreach ($semester->questionsNotAnsweredBy(user()) as $question) {
```
kdmnk: A clearer flow would be to check if the user answered or not and if not, get all questions for the semester.

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
```diff
@@ -0,0 +1,70 @@
...
+return new class () extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('answer_sheets', function (Blueprint $table) {
+            $table->id();
+            $table->unsignedSmallInteger('semester_id');
+            // $table->boolean('is_resident'); // but the BBs... let this be a question
```
kdmnk: delete comment. Since we can add these type of questions dynamically, if needed.

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
```diff
@@ -0,0 +1,70 @@
...
+            $table->unsignedSmallInteger('year_of_acceptance');
```
kdmnk, line 17: same for year_of_acceptance? Does it really needed? I feel like this just causes more trouble

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
```diff
@@ -0,0 +1,70 @@
...
+            $table->renameColumn('general_assembly_id', 'parent_id');
+            $table->boolean('has_long_answers'); // whether it expects a long text answer
+        });
+        DB::table('questions')->update(['parent_type' => 'App\Models\GeneralAssemblies\GeneralAssembly']);
+        // and then make it non-nullable
+        Schema::table('questions', function (Blueprint $table) {
+            $table->string('parent_type')->nullable(false)->change();
+        });
+
+        Schema::create('answer_sheet_question_option', function (Blueprint $table) {
```
kdmnk, line 36: The answers should be able to go to the same table for general assemblies and semester evaluations. We don't need this then

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
```diff
@@ -0,0 +1,70 @@
...
+            // no timestamps
+
+            $table->foreign('answer_sheet_id')->references('id')->on('answer_sheets');
+            $table->foreign('question_option_id')->references('id')->on('question_options');
+        });
+
+        Schema::create('long_answers', function (Blueprint $table) {
+            $table->id();
+            $table->unsignedBigInteger('question_id');
+            $table->unsignedBigInteger('answer_sheet_id');
```
kdmnk, line 48: delete answer sheet id. In the lower level, we need to support long answers for all questions (even general assemblies), even if it won't happen.

database/seeders/DatabaseSeeder.php
```diff
@@ -13,14 +13,13 @@ class DatabaseSeeder extends Seeder
      */
     public function run()
     {
-        config(['mail.driver' => 'log']);
```
kdmnk, line 16: why did you delete? This ensured that we are not sending emails during seeding

resources/views/layouts/navbar.blade.php
```diff
@@ -122,6 +122,17 @@ class="material-icons left">person_search</i>Felvételi</a></li>
                                     </a>
                                 </li>
                                 @endcan
+                                @can('is-collegist')
```
kdmnk: Unnecessary

resources/views/secretariat/evaluation-form/anonymous_questions.blade.php
```diff
@@ -0,0 +1,48 @@
...
+    @lang('anonymous_questions.all_questions_filled')
+    @else
+
+    <ul class="collection">
+        @foreach ($questions as $question)
+        <li class="collection-item">
+            <div class="question-title">{{ $question->title }}</div>
+            <div class="row">
+                @if($question->has_long_answers)
+                <x-input.textarea :id="'q'.$question->id" :text="__('anonymous_questions.long_answer_placeholder')" style="height:100px" />
```
kdmnk: Use the key getter as mentioned before

resources/views/secretariat/evaluation-form/anonymous_questions.blade.php
```diff
@@ -0,0 +1,48 @@
...
+                @else
+                @foreach($question->options()->get() as $option)
+                    @if($question->max_options==1)
+                    <x-input.radio :name="'q'.$question->id" value="{{$option->id}}" text="{{$option->title}}" />
+                    @else
+                    <x-input.checkbox :name="'q'.$question->id.'[]'" value="{{$option->id}}" text="{{$option->title}}" />
+                    @endif
+                @endforeach
+                @endif
+                @foreach ($errors->all() as $error)
```
kdmnk: move this outside of the parent foreach

kdmnk: But I believe the input components will contain the validation errors so it should not be necessary

resources/views/student-council/general-assemblies/questions/create.blade.php
```diff
@@ -1,9 +1,21 @@
...
 
+@php
+    // the parent might be a general assembly
+    // or a semester (its evaluation form)
+    $isForAssembly = isset($general_assembly)
+    // otherwise, the parent is called $semester
+@endphp
+
 @section('title')
+@if ($isForAssembly)
```
kdmnk: Have two pages for each of them. Delete all the ifs. You can make a general component for the input fields, passing a value if long answers can be enabled or not.

kdmnk (commented):
Wow, that's nice job. This is a complex problem so I added a lot of comments related how I think the morph relation should work. 2 reasons: without proper abstrabtion levels, it's just simpler to make another, similar module without combining them; this will also enable us to add anonym questions for other modules (eg. mr and miss).
Read through them and let me know if something is not clear.

app/Models/AnonymousQuestions/AnswerSheet.php
kdmnk: Okay I think I see now. Is this only to group the different questions filled by the same user? Then would not that be simpler to group them based on a random number? Anything that can not be linked to the user would do. Just store the same value for the same user for each answer, and since the data can not change, you can group them anytime later.

app/Http/Controllers/Secretariat/AnonymousQuestionController.php
kdmnk: This is for the other places where you use this function. Here just get the rules for all questions.

kdmnk:
> > Now, it's not technically mandatory to fill them; my personal opinion is that we need not explicitly tell fillers about it, as they otherwise won't fill:D But could you talk to the Ethics Commissioners since then?
> 
> I will respond to the other thread soon which probably needs to be addressed. As it has been discussed privately, I did not escalate this issue to the Ethics Commissioners, as I don't think that I would violate ethical norms by deploying this.

About these, if the user does not fill something in, you can still store an empty string or zero voted options in the database to mark the question as filled if you need to.
I also vote for not forcing users to answer about feedback.

kdmnk:
And about security issues: a proper e-voting solution would probably be way more complex and something that we can't and probably don't want to maintain. We should be proud that we have even gotten this far with this solution. 

The users should be informed of the risks of course, however, it is a small community and the system administrators have been already made responsible for keeping the user data safe. 

The users can always choose to give offline feedback, and if any issue may arise, you can always go back for paper voting (so the people who do not agree vote offline)

app/Http/Controllers/Secretariat/AnonymousQuestionController.php
viktorcsimma: The idea here is that if we add more questions while the form is still open (which we shouldn't:D), then the user will be able to fill those too when they open the form. But I don't insist on this. What do you think?

app/Http/Controllers/Secretariat/AnonymousQuestionController.php
viktorcsimma: The problem is within the unit tests; there, we must provide different users every time. Something that I can think of would be to add methods like `giveAnonymousAnswerByCurrentUser` which use user(), but I am not sure that would simplify the matter. Or am I missing something?

kdmnk: Ah, true. Then ignore this

app/Http/Controllers/Secretariat/AnonymousQuestionController.php
kdmnk: Hm... That might happen. Ok

app/Http/Controllers/StudentsCouncil/GeneralAssemblyQuestionController.php
```diff
@@ -18,7 +19,7 @@ public function create(GeneralAssembly $generalAssembly)
         $this->authorize('administer', GeneralAssembly::class);
 
         if ($generalAssembly->isClosed()) {
-            abort(401, "tried to modify a general_assembly which has been closed");
+            abort(403, "tried to modify a general_assembly which has been closed");
```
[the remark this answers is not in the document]
viktorcsimma: Are such messages in other parts of the code in the format you suggest? If not, I would keep these. @coderabbitai

app/Http/Controllers/StudentsCouncil/GeneralAssemblyQuestionController.php
```diff
@@ -72,10 +74,14 @@ public function store(Request $request, GeneralAssembly $generalAssembly)
...
      */
-    public function show(GeneralAssembly $generalAssembly, $question)
+    public function show(GeneralAssembly $generalAssembly, Question $question)
     {
         $this->authorize('viewAny', GeneralAssembly::class);
-        $question = $generalAssembly->questions()->findOrFail($question);
+        // check whether it belongs here
+        // and throw a 404 if not
+        if ($generalAssembly != $question->parent) {
+            abort(404);
```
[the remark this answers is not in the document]
viktorcsimma: I will be able to eliminate those with scope bindings.

app/Models/AnonymousQuestions/AnswerSheet.php
viktorcsimma: It also contains some anonymous information about the user. For now, it only contains the year of acceptance; but if we revise the role system, we will be able to add whether they reside in the building.

app/Models/GeneralAssemblies/Question.php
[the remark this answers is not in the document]
kdmnk: The comment is meant to be for @viktorcsimma

app/Models/AnonymousQuestions/AnswerSheet.php
kdmnk: My argument is that if we want we can ask about these information if they are relevant. But I see the point.

app/Models/GeneralAssemblies/Question.php
viktorcsimma: In the other function, we only have counters to deal with; here, we must also group answers into `AnswerSheet` instances so that we can draw further conclusions (e.g. those who answered with Yes to the first question are more likely to answer No to the second one).

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
viktorcsimma: Hm... my idea is that if we gave too many questions like this, they would be less likely to fill. So it's a convenience for them.

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
viktorcsimma: Here, we have the same problem: for general assemblies, we only need the counters; here, we would also like to get correlations.

database/seeders/DatabaseSeeder.php
viktorcsimma: Oh, I see... I thought it was a duplicate. I'll add a comment to it.

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
kdmnk: Ah, I see. Then I would push the idea that we don't need the answer sheet at all and use a random identifier to group answers. This just makes the solution too complex and specific to only the semester evaluation.

app/Http/Controllers/Secretariat/AnonymousQuestionController.php
viktorcsimma: I think I don't get it; are you suggesting adding a separate function calling `$request->validate(self::answerValidationRules($semester))`? Essentially, the array generated _is_ the set of rules.

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
viktorcsimma: I think we should simply go with accepting that; it would be important for them to see that their questions do get stored and checked. Thank you for the ideas:)

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
viktorcsimma: I see... what about a connection table? Something like `answer_sheet_long_answer`?

kdmnk: Another reason for https://github.com/EotvosCollegium/mars/pull/542/files#r1636833190 ;) 

but let's put it like this:
My priority is https://github.com/EotvosCollegium/mars/pull/542#discussion_r1636565814, if you can do it with answer sheets while keeping the code reasonably nice, I'm fine with it.

app/Http/Controllers/Secretariat/AnonymousQuestionController.php
kdmnk: I want a helper function for one question that returns the rule, so you don't have 'if's within the loop. 

Ideally I would try to have something like 
$request->validate($semester->questions->flatMap(fn ($q) => [$q->key =>  $q->rule()])

app/Models/Semester.php
viktorcsimma: Hm... the problem is that only `SemesterEvaluationController` has the trait, not individual semesters. Or do we have some other means to tell when a given semester's form has been closed?

resources/views/secretariat/evaluation-form/anonymous_questions.blade.php
viktorcsimma: They did not appear for me at the inputs, for some reason. But I moved the foreach outside.

app/Models/Semester.php
kdmnk: You should use the SemesterEvaluationController's periodic event since it is connected to that.

Technically, you can also add the trait to this one and override the '$underlyingControllerName' so that this controller will use the same periodicEvent as the SemesterEvaluationController:

```
    use HasPeriodicEvent;

    function __construct()
    {
        $this->underlyingControllerName = SemesterEvaluationController::class;
    }
```
With this, I'd say rename the property to something like $periodicEventsAttachedController

app/Http/Controllers/Secretariat/AnonymousQuestionController.php
viktorcsimma: Oh, I see... okay; I'll do that.

app/Models/AnonymousQuestions/AnswerSheet.php
viktorcsimma: Also, this way they do see that this is integrated into the system and is "smart":D so maybe they will feel more inclined to fill it.

kdmnk: but we loose authenticity that it is anonymous

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
viktorcsimma: Yes; finally I've done that. So now I have a `storeAnswers` method which expects an optional `$answerSheet` argument.

app/Models/Question.php
```diff
@@ -0,0 +1,309 @@
...
+ * @method static \Illuminate\Database\Eloquent\Builder|Question whereMaxOptions($value)
+ * @method static \Illuminate\Database\Eloquent\Builder|Question whereOpenedAt($value)
+ * @method static \Illuminate\Database\Eloquent\Builder|Question whereTitle($value)
+ * @mixin \Eloquent
+ */
+class Question extends Model
+{
+    use HasFactory;
+
+    protected $fillable = ['title', 'sitting_id', 'max_options', 'opened_at', 'closed_at', 'has_long_answers'];
```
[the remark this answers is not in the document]
viktorcsimma: Nice; it was `parent_id`. Thanks!

app/Models/Question.php
```diff
@@ -0,0 +1,309 @@
...
+     */
+    public function open(): void
+    {
+        if (!$this->parent->isOpen()) {
+            throw new Exception("tried to open question when general_assembly was not open");
+        }
+        if ($this->isOpen() || $this->isClosed()) {
+            throw new Exception("tried to open question when it has already been opened");
+        }
+        $this->update(['opened_at' => now()]);
```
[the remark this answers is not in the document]
viktorcsimma: Good one; thanks:)

app/Models/Question.php
```diff
@@ -0,0 +1,309 @@
...
+                QuestionUser::create([
+                    'question_id' => $this->id,
+                    'user_id' => $user->id,
+                ]);
+                return $this->longAnswers()->create([
+                    'answer_sheet_id' => $answerSheet->id,
+                    'text' => $text
+                ]);
+            });
+        }
```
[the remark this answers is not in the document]
viktorcsimma: I think it's needless now.

viktorcsimma:
Most of it is done:) I've finally solved the `storeAnswers` method with answer sheets; though with a bit too many ifs.

There are two other points that have remained; I'll check those tomorrow morning.

app/Http/Controllers/StudentsCouncil/GeneralAssemblyQuestionController.php
```diff
@@ -84,16 +86,15 @@ public function show(GeneralAssembly $generalAssembly, $question)
...
-    public function openQuestion(GeneralAssembly $generalAssembly, $question)
+    public function openQuestion(GeneralAssembly $generalAssembly, Question $question)
     {
         $this->authorize('administer', GeneralAssembly::class);
-        $this->authorize('administer', GeneralAssembly::class);
-        $question = $generalAssembly->questions()->findOrFail($question);
+
         if (!$generalAssembly->isOpen()) {
-            abort(401, "tried to open a question when the sitting itself was not open");
+            abort(403, "tried to open a question when the sitting itself was not open");
```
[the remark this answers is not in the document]
viktorcsimma: Thanks; I'll rephrase that.

app/Models/Semester.php
viktorcsimma: Can this return different end dates for different semesters? Or am I missing the point?

viktorcsimma: Or can we also elect to consider all questions (including those of the previous semesters) as open when the evaluation form is open? Technically, it is not ideal, but it would simplify the code greatly while also not causing practical problems (they will probably only fill the questions on the form).

app/Models/GeneralAssemblies/Question.php
viktorcsimma: Done;)

app/Models/GeneralAssemblies/Question.php
viktorcsimma: Done;) now it sets the start and end dates when we manually activate the semester evaluation form.

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
viktorcsimma: See the comment below;) I've opted for the single `storeAnswers` method.

viktorcsimma:
Okay; now it's largely done. ~But I have been trying to fix the tests for hours now, and something is not right. See the comments below. I don't know if I am missing something trivial; could you take a look at it? Thanks in advance:)~

tests/Unit/AnonymousQuestionTest.php
```diff
@@ -0,0 +1,337 @@
...
+
+        $this->assertEquals(1, $question->options->first()->votes);
+        $this->assertEquals(0, $question->options->get(1)->votes);
+        $this->assertEquals(0, $question->options->get(2)->votes);
+
+        $this->assertTrue(
+            $question->users()
+                ->where('id', $user->id)
+                ->exists()
+        );
```
viktorcsimma: It's here. The assertTrue sometimes fails and sometimes not; if it fails, it believes there are no users belonging to the question (it echoes 0). I've tried `$question->refresh()` and `$question->load('users')`, but it remains the same. The most confusing thing is its randomness; I've had a similar problem with the opening dates, but there, `$question->refresh()` solved it.

viktorcsimma: Now, I've tried
```php
DB::table('question_user')->insert(['question_id' => $this->id, 'user_id' => $user->id]);
echo DB::table('question_user')->where('question_id', $this->id)->where('user_id', $user->id)->count() . "\n";
```
~and it sometimes echoes 0. I'm clueless:D~

---

No, it was my fault; I read it wrong.

viktorcsimma:
Okay; it seems to work now. (For some reason, it was only stable if I used the DB facade directly instead of Eloquent operations. I still don't know what happened, but I'm glad I could find a workaround:D)

If you agree to the others, I think there is only one thing left; [this conversation](https://github.com/EotvosCollegium/mars/pull/542#discussion_r1636546733).

app/Models/GeneralAssemblies/Question.php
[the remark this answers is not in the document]
viktorcsimma: @kdmnk I don't understand this one; are you proposing to remove that check? Or moving it to the controller?

app/Models/Question.php
```diff
@@ -0,0 +1,281 @@
...
+                    throw new Exception("More answers given then allowed ({$this->max_options})");
+                }
+                foreach ($answer as $option) {
+                    if ($option->question_id != $this->id) {
+                        throw new Exception("Received an option which does not belong to the question");
+                    }
+                    $option->increment('votes');
+
+                    if (isset($answerSheet)) {
+                        DB::table('answer_sheet_question_option')->insert([
```
horcsinbalint, line 222: I might be misunderstanding this, but isn't it just a simple many to many relationship that could be managed by the ORM? I don't know whether it is recommended, but I would probably use `attach` or `detach` for it. https://laravel.com/docs/11.x/eloquent-relationships#updating-many-to-many-relationships

horcsinbalint: The reason behind my suggestion is that inside Controllers, it should not matter how the data is managed inside the database and what are the field / table names as it should be handled inside the model.

horcsinbalint: Oh, nvm, I think I can probably see why

app/Models/Question.php
```diff
@@ -0,0 +1,281 @@
...
+            } else {
+                $this->longAnswers()->create([
+                    'answer_sheet_id' => $answerSheet->id,
+                    'text' => $answer
+                ]);
+            }
+
+            try {
+                // For some reason, it seems to be stable
+                // only if we manipulate the database directly.
```
horcsinbalint: Isn't it because these constraints are managed by the database and Eloquent does some caching?

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
horcsinbalint: I think `// no timestamps (for better anonymity)` could be misleading. I would add `, the order in the database might break anonymity`, as it is highly likely, that no one would look back into this conversation.

horcsinbalint: It is trivial that by using a centralized for web interface voting, the voters need to trust the system administrators (it is impossible to fix this - https://youtu.be/LkH2r-sNjQs ). But system administrators should know what kind of data should be handled really carefully.

app/Models/Question.php
```diff
@@ -0,0 +1,281 @@
...
+ * @method static \Illuminate\Database\Eloquent\Builder|Question whereMaxOptions($value)
+ * @method static \Illuminate\Database\Eloquent\Builder|Question whereOpenedAt($value)
+ * @method static \Illuminate\Database\Eloquent\Builder|Question whereTitle($value)
+ * @mixin \Eloquent
+ */
+class Question extends Model
+{
+    use HasFactory;
+
+    protected $fillable = ['title', 'sitting_id', 'max_options', 'opened_at', 'closed_at', 'has_long_answers'];
```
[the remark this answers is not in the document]
viktorcsimma: True; thanks!

app/Models/Question.php
viktorcsimma: Yes; after the previous saga, I'd rather manipulate the database directly :sweat_smile:

app/Models/Question.php
viktorcsimma: It sometimes works and sometimes not; so I agree there must be some kind of caching. But I've tried almost everything I could to refresh Eloquent models, with no result.

tests/Unit/AnonymousQuestionTest.php
```diff
@@ -0,0 +1,329 @@
...
+        $this->openForm();
+        $question->refresh();
+
+        $this->expectException(\Exception::class);
+        $question->storeAnswers(
+            $user,
+            $question->options->random(2)->all(),
+            AnswerSheet::createForUser($user, $semester)
+        );
+    }
```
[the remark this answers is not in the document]
viktorcsimma: Good idea; thanks:)

app/Http/Controllers/Secretariat/SemesterEvaluationController.php
```diff
@@ -152,7 +158,7 @@ public function store(Request $request)
         $this->authorize('fill', SemesterEvaluation::class);
 
         $validator = Validator::make($request->all(), [
-            'section' => 'required|in:alfonso,courses,avg,general_assembly,feedback,other,status',
+            'section' => 'required|in:alfonso,courses,avg,general_assembly,anonymous_questions,other,status',
```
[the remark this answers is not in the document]
viktorcsimma: No; it's a name of a section in the 'secretariat.evaluation-form.app' view.

app/Models/Question.php
```diff
@@ -0,0 +1,309 @@
...
+     */
+    public function close(): void
+    {
+        if ($this->isClosed()) {
+            throw new Exception("tried to close question when it has already been closed");
+        }
+        if (!$this->hasBeenOpened()) {
+            throw new Exception("tried to close question when it has not been opened");
+        }
+        $this->update(['closed_at' => now()]);
```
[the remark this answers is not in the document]
viktorcsimma: No; it's more specific this way.

app/Models/Question.php
```diff
@@ -0,0 +1,309 @@
...
+        return $this->morphTo();
+    }
+
+    /**
+     * Whether the parent is a general assembly
+     * or a semester (evaluation form).
+     */
+    public function isForAssembly(): bool
+    {
+        return GeneralAssembly::class == $this->parent_type;
```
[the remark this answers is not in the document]
viktorcsimma: Thanks! I've already removed this method; so it's not necessary anyway.

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
horcsinbalint: As I understand, the current implementation would only provide pseudonymouzation in case of long answers and would provide proper anonymization in cases where it is aggregated during the voting.

My main takeaway here is not the video's title I linked (that electronic voting is bad). But rather that we should be aware of our system's limitations.
That is why I asked what do we call an anonymous answer.

This is not necessarily a problem.

So we could change it to `// no timestamps (for better pseudonymity)`

viktorcsimma: What about this?

```
// no timestamps (for better pseudonimity)
// note: the order of records might still break anonymity
```

horcsinbalint: Yes, I like it

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
```diff
@@ -0,0 +1,171 @@
...
+        }
+        $validator->validate();
+
+        // NOTE: for now, we set the start and end dates for the _current_ semester;
+        // they will be updated when we set the dates for the semester evaluation form
+        $question = $semester->questions()->create([
+            'title' => $request->title,
+            'max_options' => $hasLongAnswers ? 0 : $request->max_options,
+            'has_long_answers' => $hasLongAnswers,
+            'opened_at' => (app(SemesterEvaluationController::class))->getStartDate() ?? Carbon::now(),
```
kdmnk: didn't the solution I suggested to add HasPeriodicEvents here work?

app/Models/Semester.php
kdmnk: With the suggested solution, the SemesterEvaluationController and the AnonymousQuestionsController will use the same PeriodicEvent, the. one that is connected and can be updated in SemesterEvaluationController.

resources/views/utils/question_card.blade.php
```diff
@@ -0,0 +1,34 @@
...
+    </div>
+    <div class="row">
+        <x-input.text type="number" value="1" id="max_options" text="voting.max_options" required/>
+    </div>
+    @if ($canHaveLongAnswers)
+    <div class="row">
+        <x-input.checkbox s="12" name="has_long_answers" text="anonymous_questions.has_long_answers"
+            {{-- disable answer options if this is checked --}}
+            onchange="
+                document.getElementById('max_options').disabled = this.checked;
```
kdmnk: add this to a new js function

kdmnk (approved):
Looks better! And also I don't really have more time to review. Good for now if the leftover comments are resolved

kdmnk:
And good job on the tests!

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
viktorcsimma: I think it does:) What about adding a `periodicEventForSemester` method so that we can access the event for the correct semester? I've tried that; it seems to work for me.

app/Models/Question.php
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    /**
+     * Opens the question.
+     * @throws Exception if it has already been opened.
+     */
+    public function open(): void
+    {
+        if (!$this->parent->isOpen()) {
+            throw new Exception("tried to open question when general_assembly was not open");
```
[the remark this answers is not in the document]
viktorcsimma: Done:)

app/Models/Semester.php
```diff
@@ -230,11 +245,39 @@ public function transactionsInCheckout(Checkout $checkout)
...
+
+    /**
+     * The questions which a given user has not yet answered.
+     * This returns a Collection.
+     */
+    public function questionsNotAnsweredBy(User $user)
+    {
+        return $this->questions
+                ->filter(function ($question) use ($user) {return !$question->hasVoted($user);});
+    }
```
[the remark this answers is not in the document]
viktorcsimma: Done:)

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
kdmnk: It does not really make sense. We have one periodicEvent which might belong to the semester you are looking for or not. The attached semester to the periodicEvent is only there to fix the semester used by the context. You should set the semester id of the feedback based on that, if you don't have that yet.

kdmnk:
> Okay; it seems to work now. (For some reason, it was only stable if I used the DB facade directly instead of Eloquent operations. I still don't know what happened, but I'm glad I could find a workaround:D)

if you want, I can take a look tomorrow

app/Utils/HasPeriodicEvent.php
```diff
@@ -44,6 +44,21 @@ final public function periodicEvent(): ?PeriodicEvent
...
+     * @return PeriodicEvent|null
+     */
+    final public function periodicEventForSemester(?Semester $semester): ?PeriodicEvent
+    {
+        if (is_null($semester)) $semester = Semester::current();
+        return PeriodicEvent::where('event_model', $this->underlyingControllerName)
+            ->where('semester_id', $semester->id)
+            ->first();
+    }
+
```
viktorcsimma: @kdmnk that's what I thought of. And we call this method with the semester for which we are creating the question. And if that semester is in the future, this will return null, and the question will not be opened.

viktorcsimma:
> > Okay; it seems to work now. (For some reason, it was only stable if I used the DB facade directly instead of Eloquent operations. I still don't know what happened, but I'm glad I could find a workaround:D)
> 
> if you want, I can take a look tomorrow

Thank you:) but it's not important for now; the priority would be to deploy this as soon as possible.

horcsinbalint:
> Thank you:) but it's not important for now; the priority would be to deploy this as soon as possible.

Is this test finished on the staging system? https://github.com/EotvosCollegium/mars/pull/532#issuecomment-2152415400

viktorcsimma:
> > Thank you:) but it's not important for now; the priority would be to deploy this as soon as possible.
> 
> Is this test finished on the staging system? [#532 (comment)](https://github.com/EotvosCollegium/mars/pull/532#issuecomment-2152415400)

Yes; it was quite a long saga, but in the end, it got merged to development. I think you have also tested it; see #547.

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public function down(): void
+    {
+        Schema::dropIfExists('long_answers');
+
+        Schema::table('questions', function (Blueprint $table) {
+            $table->dropColumn('has_long_answers');
+            $table->dropColumn('parent_type');
+            $table->renameColumn('parent_id', 'general_assembly_id');
+        });
```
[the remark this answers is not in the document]
viktorcsimma: I think they are the other way round.

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        Schema::create('answer_sheets', function (Blueprint $table) {
+            $table->id();
+            $table->unsignedSmallInteger('semester_id');
+            $table->unsignedSmallInteger('year_of_acceptance');
+            // no timestamps (for better pseudonimity)
+            // note: the order of records might still break anonymity
+
+            $table->foreign('semester_id')->references('id')->on('semesters');
+        });
```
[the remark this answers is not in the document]
viktorcsimma: Yes; we have talked about it and it will be fine. `year_of_acceptance` is not related anyway.

horcsinbalint:
> > > Thank you:) but it's not important for now; the priority would be to deploy this as soon as possible.
> > 
> > 
> > Is this test finished on the staging system? [#532 (comment)](https://github.com/EotvosCollegium/mars/pull/532#issuecomment-2152415400)
> 
> Yes; it was quite a long saga, but in the end, it got merged to development. I think you have also tested it; see #547.

It was requested from me to deploy it to staging https://github.com/EotvosCollegium/mars/pull/532#issuecomment-2152415400 :

> @horcsinbalint please let me know when can I test it on the test server, if that's available

I have deployed the changes to the staging server and sent credentials privately. #543 and #548 are quite big and I have not received confirmation whether it works with live data and if it is safe to deploy on the live environment.

database/migrations/2024_06_01_173029_add_anonymous_questions_to_evaluation_form.php
```diff
@@ -0,0 +1,70 @@
...
+            // for a polymorphic relationship
+            $table->string('parent_type')->after('general_assembly_id')->nullable();
+            $table->renameColumn('general_assembly_id', 'parent_id');
+            $table->boolean('has_long_answers'); // whether it expects a long text answer
+        });
+        DB::table('questions')->update(['parent_type' => 'App\Models\GeneralAssemblies\GeneralAssembly']);
+        // and then make it non-nullable
+        Schema::table('questions', function (Blueprint $table) {
+            $table->string('parent_type')->nullable(false)->change();
+        });
```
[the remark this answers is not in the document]
viktorcsimma: I'll leave that to the database system to decide.

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
```diff
@@ -0,0 +1,182 @@
...
+     */
+    public function exportAnswerSheets(Semester $semester)
+    {
+        $this->authorize('administer', AnswerSheet::class);
+
+        return Excel::download(
+            new AnonymousQuestionsExport($semester),
+            'anonymous_questions_' . $semester->year . '_' . $semester->part . '.xlsx'
+        );
+    }
```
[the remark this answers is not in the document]
viktorcsimma: Only the year of acceptance is included; that will be fine. Maybe a randomisation would help.

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
```diff
@@ -0,0 +1,182 @@
...
+                );
+                $question->storeAnswers(user(), $options, $answerSheet);
+            } else {
+                $option = QuestionOption::find($answer);
+                $question->storeAnswers(user(), $option, $answerSheet);
+            }
+        }
+
+        return back()->with('message', __('general.successful_modification'))->with('section', $request->section);
+    }
```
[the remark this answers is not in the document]
viktorcsimma: It will be fine for now.

horcsinbalint:
Or at least I have not tested it (other than logging into it and catching #544). But I have no idea if the code is fit for production.

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
```diff
@@ -0,0 +1,182 @@
...
+                    'title' => $option,
+                    'votes' => 0
+                ]);
+            }
+        }
+
+        session()->put('section', $semester->id);
+        return redirect()->route('anonymous_questions.index_semesters')
+                         ->with('message', __('general.successful_modification'));
+    }
```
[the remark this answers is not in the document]
viktorcsimma: The end is not specific; so this will be better.

kdmnk:
@horcsinbalint please keep the chat here only for the PR and code related messages, it is already quite hectic.

app/Utils/HasPeriodicEvent.php
kdmnk: This won't give you any data for older semesters. 
The information flow should be: you get/set the periodicEvent, and through that, you get the semester that you can use. 

When you create a questions, you store them with the semester you get from the current periodicEvent.
Afterwards, you can still query older questions by directly using the semester ids, without the periodicEvent as it might have been updated for a new voting period.

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        $this->authorize('administer', AnswerSheet::class);
+
+        if ($semester->isClosed()) {
+            abort(403, "tried to add a question to a closed semester");
+        }
+        return view('student-council.anonymous-questions.create', [
+            "semester" => $semester
+        ]);
+    }
```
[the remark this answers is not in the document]
viktorcsimma: It will be fine for now.

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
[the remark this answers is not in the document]
viktorcsimma: It's done with the last commit:)

app/Utils/HasPeriodicEvent.php
viktorcsimma: Oh; so those are only there for querying older ones... Maybe the easiest way for now would be to only allow admins to add questions for the current semester. (They probably won't want to do so for future semesters anyway.)

Let me know what you think. If you agree, I'll proceed with that.

kdmnk: Well if you need this asap I think go back to the embedded google forms :D 

We could try to release it, but I'm trying not to release half done features anymore. I will have time during the weekend to have a deeper look again and give some solutions for our uncertainties. (Currently the morph relation's complexity overcomes its advantages, but I have to take another look).

Again, after the previous years of experience, our priority here is to produce software that is easily maintainable and understandable, which is more important to me than releasing new features.
After the discussions, I might also need to go back to the PeriodicEvents as I can see it is not clearly understandable and does not really cover this case. 

What do you think?

kdmnk: I also understand if you want to release this and will go back to it afterwards if you are confident about the current solution's behaviour.

viktorcsimma: > I also understand if you want to release this and will go back to it afterwards if you are confident about the current solution's behaviour.

Yes, that would be the best for me now:) and after the form gets closed, we can open a new issue for it. I think I can work on this more on July.

Thank you 😊

horcsinbalint: I advise against deploying changes in the codebase that are not sure to be correct. I also have a problem on an operational standpoint as in June there will not be enough human resources to restore a backup if that is needed due to a mistake presented in the newly deployed code. But I will deploy the changes if someone takes the responsibility even though I would not recommend it.

viktorcsimma: Okay; I think I've played with it long enough to say it's going to be alright. So I'll take the blame if something happens.

viktorcsimma:
Then if you are fine with it; I'll merge this now and then we can reopen discussion in the summer. And I'll take the blame if anything blows up;) Thank you both very much!

PR #551, opened 2024-06-15
Deploy recent changes

horcsinbalint:
## Description
Deploy recent changes


viktorcsimma (approved):
There were some missing backslashes before `App\Models\...` notations which caused checkouts to crash; I've added them.

Otherwise, it seems to be OK for me. The checkboxes again seem to be filled automatically, but I don't want to complicate the process any further.

horcsinbalint:
> Otherwise, it seems to be OK for me. The checkboxes again seem to be filled automatically, but I don't want to complicate the process any further.

I just usually bypass branch protections when I want to change `staging` or `deployment` as there is almost always reviewed code on `development`. :D

So is it the version I should deploy to production?

kdmnk:
Bypassing branch protections is perfectly fine. Even better if you can just merge and push locally.

PR #552, opened 2024-06-15
Quick bugfixes for anonymous questions

viktorcsimma:
Bugfixes for the anonymous questions feature introduced in #542.

- Now, the validator handles if more than `max_options` options are checked for a multiple-choice question.
- After validation errors, all checkboxes were checked and all radio buttons were unchecked. Now, they hold their values.
- After validation errors, the redirection passes the appropriate section to the session variables, so the user need not scroll down again.
- Creation of answer sheets is now wrapped into a `DB::transaction`.
- The header of the Excel sheet still had the _id_ field.
- Integrated #526 here (the text is now `A beküldési határidőig bármikor beírhatod majd.`).
- On the admin page,
  - future semesters are no longer displayed so as to avoid confusion;
  - the card for the current semester is automatically set active.

Huge thanks if you can take a look at them before deployment.

### DIFF
--- a/app/Exports/UsersSheets/AnonymousQuestionsExport.php
+++ b/app/Exports/UsersSheets/AnonymousQuestionsExport.php
@@ -53,7 +53,6 @@ class AnonymousQuestionsExport implements FromCollection, WithMapping, WithHeadi
     public function headings(): array
     {
         return array_merge([
-            __('general.id'),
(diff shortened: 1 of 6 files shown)


kdmnk (approved):

viktorcsimma:
Thank you:)

PR #553, opened 2024-06-15
Added phpstan for static analysis

kdmnk:
I just found this tool randomly, but looks nice, I will try to see if I can fix the issues.


kdmnk:
I will merge this as this only contains minor code fixes, mostly related to typing that is reported by the tool.

PR #554, opened 2024-06-15
Suggestions to use periodicEvents for AnonymousQuestions

kdmnk:
@viktorcsimma I think I know what you found confusing about the periodicEvents.

With PeriodicEvents, you cannot create a question for a previous semester, you can only create and answer a question for the current period (which makes sense to me, but let me know if you don't agree).
When the dates of the semesterEvaluation are set, the semester to which it applies must also be specified. You can use that semester via the periodicEvent which will also define the deadlines for the questions.

Questions for older semesters can still be exported as they are stored for their semesters.

This PR contains how I think the controller should utilise the PeriodicEvents. The other parts of the solution should be modified accordingly.

Feel free to take over this branch or create a new one.

### DIFF
--- a/app/Http/Controllers/Secretariat/SemesterEvaluationController.php
+++ b/app/Http/Controllers/Secretariat/SemesterEvaluationController.php
@@ -57,7 +57,7 @@ class SemesterEvaluationController extends Controller
 
         $this->updatePeriodicEvent($semester, $startDate, $endDate);
 
-        // setting start and end dates of questions
+        // updating start and end dates of questions
         $semester->questions()->update([
             'opened_at' => $startDate,
             'closed_at' => $endDate
--- a/app/Http/Controllers/Secretariat/SemesterEvaluationController.php
+++ b/app/Http/Controllers/Secretariat/SemesterEvaluationController.php
@@ -136,6 +136,9 @@ class SemesterEvaluationController extends Controller
         $this->authorize('fillOrManage', SemesterEvaluation::class);
 
         return view('secretariat.evaluation-form.app', [
+            // let the current semester be found based on the periodic event itself
+            // we can safely assume it is not null
+            'semester' => app(\App\Http\Controllers\Secretariat\SemesterEvaluationController::class)->semester(),
             'phd' => user()->educationalInformation->studyLines()->currentlyEnrolled()->where('type', 'phd')->exists(),
             'user' => user(),
             'faculties' => Faculty::all(),
--- a/app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
+++ b/app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
@@ -45,24 +45,27 @@ class AnonymousQuestionController extends Controller
     /**
      * Returns the 'new question' page.
      */
-    public function create()
+    public function create(Semester $semester)
     {
         $this->authorize('administer', AnswerSheet::class);
 
-        if (!$this->isActive()) {
+        if ($semester->isClosed() && !$semester->isCurrent()) {
             abort(403, "tried to add a question to a closed semester");
         }
-        return view('student-council.anonymous-questions.create');
+        return view('student-council.anonymous-questions.create', [
+            "semester" => $semester
+        ]);
     }
 
     /**
      * Saves a new question.
      */
-    public function store(Request $request)
+    public function store(Request $request, Semester $semester)
     {
         $this->authorize('administer', AnswerSheet::class);
 
-        if (!$this->isActive()) {
+        // we need this; the current semester might also be closed
+        if ($semester->isClosed() && !$semester->isCurrent()) {
             abort(403, "tried to add a question to a closed semester");
         }
 
@@ -86,14 +89,11 @@ class AnonymousQuestionController extends Controller
         }
         $validator->validate();
 
-        $event = $this->periodicEvent();
-
-        $question = $event->semester->questions()->create([
+        $question = $semester->questions()->create([
             'title' => $request->title,
             'max_options' => $hasLongAnswers ? 0 : $request->max_options,
             'has_long_answers' => $hasLongAnswers,
-            // does not have individual dates, voting is constrained by the periodicEvent
-            'opened_at' => null,
+            'opened_at' => \Carbon\Carbon::now(),   // let it simply be open by default
             'closed_at' => null
         ]);
         if (!$hasLongAnswers) {
@@ -106,7 +106,7 @@ class AnonymousQuestionController extends Controller
         }
 
         //session()->put('section', $semester->id);
-        return redirect()->route('anonymous_questions.index_semesters')
+        return redirect()->route('anonymous_questions.index')
                          ->with('message', __('general.successful_modification'));
     }
 
@@ -175,7 +175,7 @@ class AnonymousQuestionController extends Controller
      * Returns an Excel sheet containing all the answers
      * to the questions of a given semester.
      */
-    public function exportAnswerSheets(Semester $semester)
+    public function exportAnswers(Semester $semester)
     {
         $this->authorize('administer', AnswerSheet::class);
 
--- a/app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
+++ b/app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
@@ -48,8 +48,9 @@ class AnonymousQuestionController extends Controller
     public function canAddQuestionTo(Semester $semester): bool
     {
         // true for future, false for past semesters
-        if (!$semester->isCurrent()) return !$semester->isClosed();
-        else {
+        if (!$semester->isCurrent()) {
+            return !$semester->isClosed();
+        } else {
             $endDate = $this->getEndDate();
             return is_null($endDate) || !$endDate->isPast();
         }
--- a/app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
+++ b/app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
@@ -55,7 +55,7 @@ class AnonymousQuestionController extends Controller
             $endDate = $this->getEndDate();
             return is_null($endDate)
                 || $this->semester()->id == $semester->id && !$endDate->isPast();
-                                         // ^ it must be the semester belonging to the periodic event
+            // ^ it must be the semester belonging to the periodic event
         }
     }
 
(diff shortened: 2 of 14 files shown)


app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
```diff
@@ -37,36 +35,34 @@ public function __construct()
...
      * Returns the 'new question' page.
      */
-    public function create(Semester $semester)
+    public function create()
     {
         $this->authorize('administer', AnswerSheet::class);
 
-        if ($semester->isClosed()) {
+        if (!$this->isActive()) {
             abort(403, "tried to add a question to a closed semester");
```
kdmnk: @viktorcsimma 
actually you don't need this. Just check if $this->periodicEvent() is not null. 
Then you can create questions for the current period, regardless if it has been opened or closed. 
Just make sure that the users update the period and the semester in the semester evaluation page beforehand (or show the currently selected semester at the question form).

viktorcsimma:
Thank you:) I start to get the idea. So if I understand right, there is only at most one PeriodicEvent connected to a controller. If a new semester comes, the `semester_id` field is updated in `updatePeriodicEvent`. Still, because the things belonging to the event get the semester id when they get created, they can be retrieved later.

I think it's fine if they can only create questions for the current semester; I agree it isn't worth the fuss to allow them to manipulate future semesters which are half a year later. I just didn't understand then.

Thank you again; I think I'll pick it up here in the summer;)

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
viktorcsimma: I see... this seems to be fine.

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
```diff
@@ -90,14 +86,15 @@ public function store(Request $request, Semester $semester)
...
-        $question = $semester->questions()->create([
+        $question = $event->semester->questions()->create([
             'title' => $request->title,
             'max_options' => $hasLongAnswers ? 0 : $request->max_options,
             'has_long_answers' => $hasLongAnswers,
-            'opened_at' => $event?->start_date ?? null,
-            'closed_at' => $event?->end_date ?? null
+            // does not have individual dates, voting is constrained by the periodicEvent
+            'opened_at' => null,
+            'closed_at' => null
```
viktorcsimma: I'm just wondering whether we could use these dates for something useful. But I don't have any ideas; resolve this if neither do you.

viktorcsimma (commented):
I'm really starting to get the idea; thank you:) I'll take over from here.

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
```diff
@@ -37,36 +35,34 @@ public function __construct()
...
         $this->authorize('administer', AnswerSheet::class);
 
-        if ($semester->isClosed()) {
+        if (!$this->isActive()) {
             abort(403, "tried to add a question to a closed semester");
         }
-        return view('student-council.anonymous-questions.create', [
-            "semester" => $semester
-        ]);
+        return view('student-council.anonymous-questions.create');
```
viktorcsimma, line 73: This here should remain in the backend, I think. We could choose it in the Blade file, too, but it would make that somewhat less general, and this is where it would be expected anyway.

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
```diff
@@ -37,36 +35,34 @@ public function __construct()
...
     /**
      * Returns the 'new question' page.
      */
-    public function create(Semester $semester)
+    public function create()
     {
         $this->authorize('administer', AnswerSheet::class);
 
-        if ($semester->isClosed()) {
+        if (!$this->isActive()) {
```
viktorcsimma: I think we do need the semester here. Otherwise, we cannot reliably determine for which semester we should create the question (especially if no periodic event exists yet).

kdmnk: We set the semester for the periodic event that you can use (See the store function, I'm getting the periodic Event and setting the question for that semester). We should not use the dates for the semesters as they are not reliable - see our original issue before the periodic events got introduced.

What do you want to achieve?

viktorcsimma: We need to determine on the index page which semester is the currently active one. If only the one with an open periodic event is allowed to be added questions, the user cannot add a question to a semester whose evaluation form has not yet been opened. So they have to open it first and _then_ add the question, which is not ideal.

What if we allow every semester which is the current one or is in the future? This only affects the council, not ordinary collegists.

resources/views/student-council/anonymous-questions/index.blade.php
```diff
@@ -45,7 +45,7 @@ class="right" :text="__('anonymous_questions.export')" />
                 @endforeach
             </ul>
 
-            @if(!$semester->isClosed())
+            @if($semester->isCurrent() || !$semester->isClosed())
             <div class="row" style="margin: 0">
                 <x-input.button :href="route('anonymous_questions.create', $semester)"
                         class="right green" :text="__('anonymous_questions.create_question')" />
```
viktorcsimma: This is the idea. That way, the council could add questions to the current and future semesters, but collegists would not see anything from that. There, we can simply proceed as you've suggested; the appropriate semester can easily be determined for an open evaluation form.

resources/views/secretariat/evaluation-form/anonymous_questions.blade.php
```diff
@@ -1,12 +1,19 @@
...
 </blockquote>
-<form method="POST" action="{{ route('anonymous_questions.store_answer_sheet', App\Models\Semester::current()) }}">
+
+@php
+// let the current semester be found based on the periodic event itself
+// we can safely assume it is not null
+$semester = app(\App\Http\Controllers\Secretariat\SemesterEvaluationController::class)->semester();
+@endphp
+
+<form method="POST" action="{{ route('anonymous_questions.store_answers', $semester) }}">
```
viktorcsimma, line 5: Here it is; here, we really use the periodic event to find the appropriate semester.

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
kdmnk: I see. So instead of checking that the periodic event is active, you can check $periodicEvent->endDate()->isPast() and allow future forms to be modified.

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
```diff
@@ -37,11 +35,25 @@ public function __construct()
...
+     */
+    public function canAddQuestionTo(Semester $semester): bool
+    {
+        // true for future, false for past semesters
+        if (!$semester->isCurrent()) {
+            return !$semester->isClosed();
+        } else {
+            $endDate = $this->getEndDate();
+            return is_null($endDate) || !$endDate->isPast();
+        }
```
viktorcsimma, line 62: What about this?

resources/views/secretariat/evaluation-form/anonymous_questions.blade.php
kdmnk: Yes but completely useless (and ugly) :D . You get the semester in the blade and pass it in the argument. Then it's equivalent to just get the semester within the controller function.

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
kdmnk: Creating a new question should create it to the current semester evaluation that is set based on the periodic event (regardless if it is active or not). In the UI, make it clear that the questions are going to be created to the semester that is selected on the periodic event. 
And you don't need this policy.

You don't need to map any post request to a specific semester. 
The older semester's results can be viewed, that's all.

I'm sorry if it is still confusing for you, maybe we can have a meeting to discuss this further and identify what should we do to make this clearer in the future. 
We finally got rid of the semesters, so we need to stick with it now.

viktorcsimma: Am I right that this means some kind of simplification so that we do not need to refer to semesters? I don't insist on being able to edit every future semester; so if you would like, I can point that out in the UI as you've suggested.

kdmnk: Yes, that's what I'm trying to say

viktorcsimma: Okay:) I'll move forward with it.

app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
```diff
@@ -43,32 +45,44 @@ public function index()
...
+    }
+
     /**
      * Returns the 'new question' page.
      */
-    public function create(Semester $semester)
+    public function create()
     {
         $this->authorize('administer', AnswerSheet::class);
+        $this->checkPeriodicEvent();
```
viktorcsimma, line 71: What about this?

resources/views/secretariat/evaluation-form/anonymous_questions.blade.php
viktorcsimma: Done:)

app/Http/Controllers/Secretariat/SemesterEvaluationController.php
```diff
@@ -136,6 +136,9 @@ public function show()
         $this->authorize('fillOrManage', SemesterEvaluation::class);
 
         return view('secretariat.evaluation-form.app', [
+            // let the current semester be found based on the periodic event itself
+            // we can safely assume it is not null
+            'semester' => app(\App\Http\Controllers\Secretariat\SemesterEvaluationController::class)->semester(),
```
kdmnk, line 141: You are already in the controlller. Use $this->semester().

resources/views/student-council/anonymous-questions/index.blade.php
```diff
@@ -5,16 +5,39 @@
 @endsection
 @section('secretariat_module') active @endsection
 
+@php
+// let the current semester be found based on the periodic event itself
+// beware: it might be null
+$periodicEvent = app(\App\Http\Controllers\Secretariat\SemesterEvaluationController::class)->periodicEvent();
```
kdmnk, line 11: Can you pass this to the blade from the controller?

PR #555, opened 2024-06-15
Decouple periodic events from controller names

kdmnk:
Now, the controllers using a periodicEvent can set which periodicEvent do they connected to.

The Controllers that use a periodicEvents can define which periodicEvent they want to edit and use in their constructor. The possible options can be found in the PeriodicEvents.
When we handle the start/end dates, we fire events and use event handlers.

Note: I had to move some functions out from the controllers so that the event handlers can take care of the functionalities.


app/Models/PeriodicEvent.php
```diff
@@ -112,13 +127,19 @@
...
     {
-        //Get the corresponding controller and call its start method
-        app($this->event_model)->handlePeriodicEventStart();
+        switch ($this->event_model) {
+            case self::SEMESTER_EVALUATION_PERIOD:
+                event(new SemesterEvaluationPeriodStart($this));
+                break;
+            default:
+                Log::debug("No event handler defined for ". $this->event_model . "'s start time");
+        }
```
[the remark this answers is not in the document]
kdmnk: Can you give an example solution?

app/Listeners/SemesterEvaluationPeriodStartListener.php
```diff
@@ -0,0 +1,28 @@
...
+     */
+    public function handle(SemesterEvaluationPeriodStart $event): void
+    {
+        Mail::to(config('contacts.mail_membra'))->queue(new EvaluationFormAvailable($event->periodicEvent->deadline()));
+        if (User::secretary()) {
+            Mail::to(User::secretary())->queue(new EvaluationFormAvailableDetails(User::secretary()->name, $event->periodicEvent->deadline()));
+        }
+        if (User::president()) {
+            Mail::to(User::president())->queue(new EvaluationFormAvailableDetails(User::president()->name, $event->periodicEvent->deadline()));
+        }
```
viktorcsimma, line 29: Should we add the director for this, too? He appears in the end listener.

app/Models/PeriodicEvent.php
```diff
@@ -112,13 +121,19 @@ public function isExtended(): bool
...
      * Handle the start of the PeriodicEvent.
      */
     public function handleStart(): void
     {
-        //Get the corresponding controller and call its start method
-        app($this->event_model)->handlePeriodicEventStart();
+        switch ($this->event_name) {
+            case self::SEMESTER_EVALUATION_PERIOD:
+                event(new SemesterEvaluationPeriodStart($this));
+                break;
```
viktorcsimma, line 133: This seems not to be open-closed; can we circumvent this somehow? Maybe through singletons derived from this class.

app/Models/PeriodicEvent.php
```diff
@@ -141,10 +161,13 @@ public function handleEnd(): void
...
     {
-        $days_left = (int)$this->endDate()->diffInDays(now()) * (-1);
-
-        //Get the corresponding controller and call its start method
-        app($this->event_model)->handlePeriodicEventReminder($days_left);
+        switch ($this->event_name) {
+            case self::SEMESTER_EVALUATION_PERIOD:
+                event(new SemesterEvaluationPeriodReminder($this));
+                break;
+            default:
```
viktorcsimma, line 168: Same here.

app/Utils/PeriodicEventController.php
```diff
@@ -56,20 +65,20 @@ final public function periodicEventForSemester(?Semester $semester): ?PeriodicEv
...
-     * @return PeriodicEvent
-     */
+    * Create or update the current PeriodicEvent connected to the model.
+    * Make sure the $data is properly validated:
+    * @param Semester $semester
+    * @param Carbon $start_date
+    * @param Carbon $end_date
+    * @param Carbon|null $extended_end_date
+    * @return PeriodicEvent
+    */
```
viktorcsimma, line 81: Nitpick, but I think the asterisks were indented right.

app/Http/Controllers/Secretariat/SemesterEvaluationController.php
```diff
@@ -141,20 +74,20 @@ public function show()
             'user' => user(),
             'faculties' => Faculty::all(),
             'workshops' => Workshop::all(),
-            'evaluation' => user()->semesterEvaluations()->where('semester_id', Semester::current()->id)->first(),
+            'evaluation' => user()->semesterEvaluations()->where('semester_id', self::semester()->id)->first(),
```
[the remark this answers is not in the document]
viktorcsimma: @kdmnk this is indeed true; what about this?
```php
where('semester_id', $this->periodicEvent()?->semester()->id ?? Semester::current())
```

app/Http/Controllers/Secretariat/SemesterEvaluationController.php
```diff
@@ -141,20 +74,20 @@ public function show()
...
             'general_assemblies' => GeneralAssembly::all()->sortByDesc('closed_at')->take(2),
-            'community_services' => user()->communityServiceRequests()->where('semester_id', Semester::current()->id)->get(),
+            'community_services' => user()->communityServiceRequests()->where('semester_id', self::semester()->id)->get(),
```
viktorcsimma: Same here.

app/Http/Controllers/Secretariat/SemesterEvaluationController.php
```diff
@@ -141,20 +74,20 @@ public function show()
...
             'position_roles' => user()->roles()->whereIn('name', Role::STUDENT_POSTION_ROLES)->get(),
             'periodicEvent' => $this->periodicEvent(),
-            'users_havent_filled_out' => user()->can('manage', SemesterEvaluation::class) ? $this->usersHaventFilledOutTheForm() : null,
+            'users_havent_filled_out' => user()->can('manage', SemesterEvaluation::class) ? User::doesntHaveStatusFor(self::semester()->succ())->get() : null,
```
viktorcsimma: And here:
```php
($this->periodicEvent()?->semester()->id ?? Semester::current())->succ()
```

viktorcsimma (changes requested):
Sorry for being so late with reviewing this. Seems to be a nice change; more understandable, clear and useful:)

Besides the things I have written (for me, the semester evaluation page crashed because of `self::semester()`, there was one more thing, seen when I switched to another periodic event:

grafik

It's only 21:48, but it is still open. Maybe my fix was incorrect.

Also, can you check out CodeRabbit comments as well?

app/Models/PeriodicEvent.php
kdmnk: Yeah I remember that I was trying a lot of stuff here. We can make this more fancier but I don't see the point of making this more complex, and with registers or whatever else we still need to do add some code manually (it was calculated automatically before, but now the point of this is to decouple the names from controllers, so the mapping has to be manual somewhere), just in a more abstracted, harder to find place. 

But I'm curious to see if you find a better way to structure this, or move it to somewhere that makes more sense to you.

app/Http/Controllers/Secretariat/SemesterEvaluationController.php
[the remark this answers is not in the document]
kdmnk: self::semester() is defined in parent class. I made it nullable though. Semester::current() might introduce confusion again.

kdmnk:
Sorry for this that it took so long. Now it also supports the deactivation through the command line that was introduced in https://github.com/EotvosCollegium/mars/pull/681. 

However, I have a docker bug now so I can not test it at the moment manually.

PR #556, opened 2024-06-15
Merge bugfixes to staging

viktorcsimma:
This has the same content as #552; it would bring those changes to staging.

### DIFF
--- a/app/Exports/UsersSheets/AnonymousQuestionsExport.php
+++ b/app/Exports/UsersSheets/AnonymousQuestionsExport.php
@@ -53,7 +53,6 @@ class AnonymousQuestionsExport implements FromCollection, WithMapping, WithHeadi
     public function headings(): array
     {
         return array_merge([
-            __('general.id'),
             __('general.semester'),
             __('user.year_of_acceptance')
         ], $this->semester->questions()->orderBy('id')->pluck('title')->all());
--- a/app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
+++ b/app/Http/Controllers/StudentsCouncil/AnonymousQuestionController.php
@@ -134,36 +134,50 @@ class AnonymousQuestionController extends Controller
     {
         $this->authorize('is-collegist');
 
-        $validatedData = $request->validate(
+        $validator = Validator::make(
+            $request->all(),
             $semester->questionsNotAnsweredBy(user())
                      ->flatMap(fn ($q) => $q->validationRules())
                      ->all()
         );
 
-        // Since answer sheets are anonymous,
-        // we cannot append new answers to the previous sheet (if any);
-        // we have to create a new one.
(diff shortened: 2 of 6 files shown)


kdmnk:
Can we reduce spam and PRs related to staging and release? :/

kdmnk (approved):

viktorcsimma:
I'm looking for solutions... for now, I've set the marscodeowners team [not to receive notifications if one of its members is also mentioned](https://github.com/orgs/EotvosCollegium/teams/marscodeowners/edit/review_assignment); but I've found nothing else.

Actually, I'm wondering whether we need a separate `staging` branch or the staging server could simply use `development`. We could talk about this later in the summer.

horcsinbalint:
> I'm looking for solutions... for now, I've set the marscodeowners team [not to receive notifications if one of its members is also mentioned](https://github.com/orgs/EotvosCollegium/teams/marscodeowners/edit/review_assignment); but I've found nothing else.

Setting code owners was only introduced for security reasons because there was a request to give out write permissions to members who are / were not system administrators. It is no longer required so we can simply remove it.

> Actually, I'm wondering whether we need a separate `staging` branch or the staging server could simply use `development`.

Yes, we need it in my experience.

horcsinbalint:
I don't really understand the problem with having a `staging` branch. It is a simple tag that helps identifying what should be on the staging system. It does not cost anything. Yes, we need to maintain this as a tag but I don't think it is equivalent to the `development` tag.

Code on `staging` should be a candidate to be released on the live system, staging is **not** a test environment. _Or at least I designed it in this way:_ https://umbraco.com/knowledge-base/staging-environment/ It is much more resource intensive to deploy changes to a staging environment than to a testing environment. (On someone's machine, updating the code is just a couple seconds, updating the staging or live environment takes up to 5-20 minutes.)

There are multiple reasons why it is done in this way:

- partial separation between how the code is developed and operated:
  - I don't want to interfere with the `development` branch if I want to test something out (if a new feature breaks my tests, I don't want to rollback the `development` branch)
  - I don't want the `development` branch to interfere with my tests (I would like to develop the infrastructure knowing what worked previously)
  - `staging` is the branch where new features are handed over from development to operations. There have been times where newer features failed during this time. If a change is not accepted, we might still need a stable `staging` code (we can achieve it by rolling back the `staging` branch to an older version).
  - version numbering is added to the `staging` branch (to know when the new features are deployed there), and it would make difficult for everyone if the version number was updated on the `development` branch.
- continuous deployment (it is under maintenance):
  - not every small change needs to be checked individually on the staging system
  - if multiple PRs are accepted in a small time-frame, what the system should do? Deployment is a resource intensive

For the last half year, I mostly did the hand over between development and operations. The current procedure worked well when there was little time pressure. (With time pressure, CD would need to be working. I did not know about plans on merging a major PR this or the next week.)

PR #562, opened 2024-06-22
Fixing some security vulnerabilities

### LINKED ISSUE #563: It is possible to inject Latex code to documents
documents/status-cert/download returns with 500 - server error

### LINKED ISSUE #567: Dangerous files can be uploaded
There are several places where it is stated that only `png` and `jpg` can be uploaded. But in reality sometimes `svg` files are also allowed. With proper operation (setting correct CSP), there is no issue. But allowing `svg` files to be uploaded can be used for an XSS attack if there are other operational errors.

horcsinbalint:
## Description

Several security vulnerabilities have been found: 

- an XSS vulnerability due to the fact that users can post SVGs to several endpoints
- most probably an RCE vulnerability as users were able to inject arbitrary latex code in several places that is processed by `pdflatex`

## Additional Notes

Our running instance has security measures which prevent these issues from being exploited.

### DIFF
--- a/app/Http/Controllers/Secretariat/UserController.php
+++ b/app/Http/Controllers/Secretariat/UserController.php
@@ -60,7 +60,7 @@ class UserController extends Controller
         session()->put('section', 'profile_picture');
 
         $request->validate([
-            'picture' => 'required|mimes:jpg,jpeg,png,gif,svg',
+            'picture' => 'required|mimes:jpg,jpeg,png,gif',
         ]);
         $path = $request->file('picture')->store('avatars');
         $old_profile = $user->profilePicture;
--- a/app/Utils/CheckoutHandler.php
+++ b/app/Utils/CheckoutHandler.php
@@ -195,7 +195,7 @@ trait CheckoutHandler
             'comment' => 'required|string',
             'amount' => 'required|integer|min:0',
             'payer' => 'exists:users,id',
-            'receipt' => 'required|mimes:pdf,jpg,jpeg,png,gif,svg',
+            'receipt' => 'required|mimes:pdf,jpg,jpeg,png,gif',
         ]);
         $validator->validate();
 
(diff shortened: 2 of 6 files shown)


kdmnk:
Nice. Can you find a general solution for this?
For example:
- Simplest: querying all data within the controller and sanitising all there
- View composers, if appliable here: https://www.slingacademy.com/article/view-composer-in-laravel-a-practical-guide/
- Middlewares, if appliable here
- etc. My goal is to be able to define new latex views while worrying as less as possible about sanitisation (see: html blades are sanitised automatically)

horcsinbalint:
Making sure that every data that is passed to the view is escaped / sanitized seems quite a bit of a pain. Like:

```blade
@foreach($items as $item)
    \item {{ $item->name }} @if(isset($item->serial_number)) ({{ $item->serial_number }}) @endif
@endforeach
```

And we would also need to be sure not to query anything inside blades, like:

```blade
E-mail: {{config("mail.secretary_mail")}} • {{\App\Models\User::director()?->email }}
```

With my proposed solution, it is at least obvious that with Latex, we should handle data carefully.
So I don't think it would be easier to do these in the controllers, view composers or middlewares. (I would definitely not know whether some data has been already escaped or not.)
So I would probably prefer having a fully manual or a fully automatic solution.

We could probably define something similar to a blade rendering engine (seems doable, but I am not sure, how laravel's internals work), or a separate, short directive like `@latex(...)` that does the escaping for latex.
But I think it is a different kind of beast compared to a hot-fix of a vulnerability. The current workaround is to remove the latex compiler from the server which causes some functionality to not being available. So I would like to bring the functionality back online if the changes fix the issue.

Therefore, I would probably prefer opening a separate issue for developing a proper long-term solution.

horcsinbalint:

kdmnk:
I'm sorry to hear. I can investigate the general solution but nowadays. I'm okay with creating a new ticket for this.

horcsinbalint:
Then, I am merging this now. I opened an issue for the generalized solution.

PR #564, opened 2024-07-10
Fixing checkbox not remembering its previous state; as well as too short question titles

viktorcsimma:
As the title says.

The solution for the anonymous questions card is not too elegant (probably the custom id causes some complications), but it seems to work for now.

### DIFF
--- a/resources/views/components/input/checkbox.blade.php
+++ b/resources/views/components/input/checkbox.blade.php
@@ -8,7 +8,8 @@
             'class' => "filled-in checkbox-color",
             'name' => $id
         ])}}
-        @checked(old($id) || $checked)
+        @checked(in_array($attributes['value'], (old($id)['option'] ?? []))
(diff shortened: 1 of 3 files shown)


resources/views/components/input/checkbox.blade.php
```diff
@@ -8,7 +8,8 @@
             'class' => "filled-in checkbox-color",
             'name' => $id
         ])}}
-        @checked(old($id) || $checked)
+        @checked(in_array($attributes['value'], (old($id)['option'] ?? []))
```
kdmnk, line 11: Have you tested this with the other checkbox instances?

viktorcsimma: Yes; with the "remember me" field and voting, it seems to work.

kdmnk (approved):

PR #566, opened 2024-07-11
Roles with time constraints

### LINKED ISSUE #392: Restructure Roles
> I was also thinking on making the roles time constrained. That way, we could have regular collegists, plus an additional resident role that is set to be valid for 1 year/indefinite. We can also use that for guests, instead of the guest_until attribute and solve #291. (Also, the general collegist role is also valid only until the end of their studies by default which we can set this way).
> 
> @machiato32 what do you think, does putting effort into this would worth it? (We have some time to work on this, the earliest this will be relevant is next summer)

> I'm starting to like this idea. The role_user table would have a created_at and valid_until field, and we would use a global scope to filter on those. This way, we don't delete roles just invalidate them, making a log out of the table.

_Originally posted by @kdmnk in https://github.com/EotvosCollegium/mars/issues/358#issuecomment-1873828995_

### LINKED ISSUE #482: Log the end of collegist status

viktorcsimma:
Closes #392. Thanks to @kdmnk for the idea:)

## Changes
- Adds the columns `created_at` and `valid_until` to the `role_users` table.
- Changes relationships so as to only include currently valid roles.
- Detaching a role now actually means setting the `valid_until` field to the current date. This might also close #482.
- Replaces the tenant role and the "resident" and "extern" role objects with a simple _resident_ role:
  - Permanent residents have both a collegist and an indefinite resident role.
  - "Resident-externs" have a collegist role as well as a resident role that has an expiry date.
    - But essentially, the same applies to externs living here as guests, just like in the old system. So we do not actually hard-wire the role into the system.
  - Externs simply have a collegist role with no resident role.
  - Tenants have a resident role with no collegist role.


database/migrations/2020_10_18_200146_set_collegist_status.php
```diff
@@ -14,7 +14,9 @@ class SetCollegistStatus extends Migration
      */
     public function up()
     {
-        $collegists = User::collegists();
+        // we have to obtain these in a legacy way,
+        // as we do not have 'valid_from' and 'valid_until' columns yet
+        $collegists = User::withRole(Role::collegist(), includesExpired: true);
```
BertalanD: This looks very troubling to me.

Are we trying to apply the current models on a database whose schema is quite different from what we use now? Should we maybe stop doing that and just interact with the database directly? IIRC we had already run into issues with migrations when deploying the previous update.

database/migrations/2024_07_10_192435_add_expiry_dates_to_roles.php
```diff
@@ -0,0 +1,131 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+use Carbon\Carbon;
+use Illuminate\Support\Facades\DB;
+use App\Models\User;
+use App\Models\Role;
```
BertalanD: Same here. Models shouldn't be referenced from migrations.

database/migrations/2020_10_18_200146_set_collegist_status.php
viktorcsimma: You are right; I'll change that:)

viktorcsimma: Actually, it turned out that this migration does not do nothing:D it does not even once enter the foreach loop, where it would call a non-existent function. So I simply deleted it.

PR #560, opened 2024-06-19
Refactor applications

kdmnk:
## Description

This is manly just to split the application and application handling part of previous ApplicationController into a new AdmissionController to prepare for #332

Other changes:
- removed statuses, instead now we only have "submitted" boolean
- detached applications from Collegist roles, now it has 'applied_for_resident_status' boolean
- found some bugs mostly in the UI
- commented out finalise functions, will fix it later
- refactored some code for better readability

More changes will come in a later PR.

PR #570, opened 2024-07-16
Fixing erroneous call which caused the tenant update panel to crash

viktorcsimma:
As in the title. It would be pretty urgent, as the tenants who have previously registered cannot login again. Could you check it out and deploy it ASAP?

### DIFF
--- a/resources/views/user/update_tenant_status.blade.php
+++ b/resources/views/user/update_tenant_status.blade.php
@@ -27,7 +27,7 @@
     </div>
 
     @if (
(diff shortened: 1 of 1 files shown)


BertalanD:
isn't this how the new API is supposed to be used instead?

```php
app(App\Http\Controllers\Auth\ApplicationController::class)->isActive()
```

See Dominik's fix in #560
https://github.com/EotvosCollegium/mars/blob/f120709f7f36e34e0c154267510714f3c1f8952d/resources/views/user/update_tenant_status.blade.php#L30

BertalanD (approved):

PR #571, opened 2024-07-16
Fixing permission problems with applications

viktorcsimma:
The problem was that old names of policies were used instead of the new ones in ApplicationFormPolicy. Therefore, applications did not appear correctly and unfinished applications were hidden.

### DIFF
--- a/app/Http/Controllers/Auth/ApplicationController.php
+++ b/app/Http/Controllers/Auth/ApplicationController.php
@@ -163,7 +163,7 @@ class ApplicationController extends Controller
                 $applications->whereIn('workshop_id', $workshops->pluck('id'));
             }
             //hide unfinished
-            if ($authUser->cannot('viewUnfinishedApplications', [User::class])) {
+            if ($authUser->cannot('viewUnfinished', \App\Models\ApplicationForm::class)) {
                 $applications->where(function ($query) {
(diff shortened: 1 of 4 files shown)


BertalanD:
(I don't know anything about PHP...) Is there a static analyzer/type checker that could catch problems like this one before broken code is deployed?

@kdmnk maybe?

kdmnk:
I have this in progress: https://github.com/EotvosCollegium/mars/pull/553
But I doubt that phpstan is checking laravel specific behavior. We are also improving our test coverage which will help. Otherwise, no, sadly.

kdmnk (approved):
LGTM, I'm just wondering what did we merge to get into this state

horcsinbalint:
> I'm just wondering what did we merge to get into this state

AFAIK the problematic code was introduced in #532 which was deployed to production by #551 because #542 depended on it (although I advised against deploying it in https://github.com/EotvosCollegium/mars/pull/542#discussion_r1639875881).

kdmnk:
> I have this in progress: #553 But I doubt that phpstan is checking laravel specific behavior. We are also improving our test coverage which will help. Otherwise, no, sadly.

Update: I found https://github.com/larastan/larastan which allegedly checks Laravel's "beautiful magic" 😄

PR #572, opened 2024-07-18
Fixing bugs in the application process prior to opening the form

viktorcsimma:
In the end, I started it based on `development`; some things have changed in it that I needed.

Largely went through @horcsinbalint's suggestions on 4th July:

```
Sziasztok!
Gyorsan próbáltam átfutni az Urán felvételivel kapcsolatos kódját. Nem voltam túl részletes, néhány probléma, amit találtam:
- app/Http/Controllers/Auth/RegisterController.php:53 rossz
- resource/views/user/update_tenant_status.blade.php:30 rossz
- felvételiztetőknek nem listázza a felvételizőket
- a határidő és a lezárás néhány dolognál csak a user interface-en látható, a POST requesteket kezelő kód nem minden esetben ellenőrzi a határidőt (profilképpel, személyes adatokkal, tanulmányi adatokkal, ALFONSÓ-val, nyelvvizsgával kapcsolatos dolgokat)
- akinek jogosultsága van egy adott felhasználót megtekinteni, annak jogosultsága van módosítani az adatokat is. Vagyis felvételiztetők módosíthatják felvételizők előbb említett adatait
- hosszú inputokat az adatbázis utasít el, egy 500-as kód társaságában, nincs validálva az input (pl.: app/Http/Controllers/Auth/ApplicationController.php:274)
- pályázónak profilképével a CSP miatt nincs lehetősége XSS-t végrehajtani, ha SVG-t tölt fel, ha jól tudom
- ha létezik már a felhasználó által meghatározott neptun kódú felhasználónak WiFije, akkor az app/Http/Controllers/Auth/ApplicationController.php:330 egy gyönyörű errort dob, és nem kap internethozzáférést az illető
- app/Http/Controllers/Auth/ApplicationController.php:317-318 sorrendje nem túl logikus olyan szempontból, hogy előbb a DB-ből a referenciát lenne talán célszerűbb kiszedni, mert így ha rosszkor száll el a rendszer, a hivatkozás megmaradhat egy nem létező fájlra
[...]

PR #574, opened 2024-07-19
Separate application workshops

### LINKED ISSUE #321: (In progress) applications without workshops do not get listed

### LINKED ISSUE #489: Separate workshop table for applicants
create a new table, with one foreign key to the original workshops and one to the application_forms. This intermediate table will store which applicant wants to get admitted to which workshop(s). This can also be used for reapplying collegists and we can  also filter on this table which workshop admits a user at the end. Make sure that the workshop leaders and the other's still access the user. This also involves separating the endpoints for updating the workshops.

kdmnk:
New ApplicationWorkshop table, so that applications can have a called in / admitted status for each workshop separately.
Also fixed viewing applicants.

Based on refactor_applications branch, so merge that first.


resources/views/auth/admission/index.blade.php
```diff
@@ -15,20 +15,20 @@
...
-                                <span style="padding-left: 25px; margin: 5px">Véglegesített</span>
+                                <input type="checkbox" name="show_not_submitted" {{$show_not_submitted ? "checked": ""}}>
+                                <span style="padding-left: 25px; margin: 5px">Nem véglegesítettek</span>
                             </label>
                         @endif
                         <x-input.button type="submit" text="Szűrés"/>
                     </form>
                     <form id="empty-filter" method="GET" action="{{route('admission.applicants.index')}}"
-                          style="{{($status=='' && $workshop=='')?'display: none':''}}">
+                          style="{{($workshop=='')?'display: none':''}}">
```
viktorcsimma, line 25: I don't understand this one. Shouldn't an admin see all applications, even ones that have no workshops yet?

resources/views/auth/application/application.blade.php
```diff
@@ -11,11 +11,64 @@
...
+                        @if ($user->application->applied_for_resident_status)
+                            <span class="new badge scale-transition coli blue tag" style="display: block;"
+                                  data-badge-caption="">
+                                @lang('role.resident')
+                            </span>
+                        @else
+                            <span class="new badge scale-transition coli orange tag" style="display: block;"
+                                  data-badge-caption="">
+                            @lang('role.extern')
+                        </span>
```
viktorcsimma: An indentation error here.

resources/views/auth/application/application.blade.php
```diff
@@ -11,11 +11,64 @@
...
+                        @endif
+                            <div class="divider" style="margin: 3px"></div>
```
viktorcsimma: There could be a greater margin around the divider.

resources/views/auth/application/application.blade.php
```diff
@@ -11,11 +11,64 @@
...
+                                                felvéve
+                                            @elseif($workshop->called_in)
+                                                behívva
+                                            @else
+                                                nincs behívva
+                                            @endif
+                                        </label>
+                                    @endif
+                                @endcan
+                            </div>
```
viktorcsimma: I think this should be omitted if the application has not yet been submitted. That would be an additional indication that that one is in progress.

resources/views/auth/application/application.blade.php
```diff
@@ -11,11 +11,64 @@
...
 
-                    <div class="card-title">{{ $user->name }}</div>
+                    <div class="right" style="margin:5px;width:150px">
+                        @can('viewUnfinished', \App\Models\Application::class)
+                            @if(!$user->application->submitted)
+                                <span class="new badge scale-transition red tag" style="display: block;"
+                                      data-badge-caption="">
+                                Nincs véglegesítve
+                                </span>
+                            @endif
```
viktorcsimma: For me, it is a bit too close to the other badges; maybe we should give a larger margin to it, or an additional divider. But this is probably a question of taste.

resources/views/auth/application/status_update_component.blade.php
```diff
@@ -1,19 +1,18 @@
...
+<div style="display: flex;flex-direction: column;align-items: center;justify-content: center;">
+    <label>
+        <input type="checkbox"
+               wire:click="callIn('{{$workshop->workshop_id}}')" @checked($workshop->called_in) />
+        <span>Behívva</span>
+    </label>
+    <label>
+        <input type="checkbox"
+               wire:click="admit('{{$workshop->workshop_id}}')" @checked($workshop->admitted) />
+        <span>Felvéve</span>
```
viktorcsimma, line 10: I think these should be radio buttons; it's strange that we can admit students without calling them in. But I don't know what to call the option before 'called in'; so we can leave it as it is if we have no better idea.

resources/views/auth/application/questions.blade.php
```diff
@@ -66,8 +65,28 @@
...
+                                <x-input.checkbox only_input id="workshop_{{$workshop->id}}" :text="$workshop->name" name="workshop[]"
+                                                  value="{{ $workshop->id }}" checked='{{$checked}}'/>
+                            </div>
+                        @endforeach
+                        </div>
+                        @error('workshop')
+                        <blockquote class="error">@lang('user.workshop_must_be_filled')</blockquote>
+                        @enderror
+                        <blockquote>
+                        Kérjük, jelentkezését csak a szakmailag releváns műhelyekbe adja le. A műhelyek egymástól függetlenül dönthetnek a meghallgatásáról.
```
viktorcsimma: I think this could imply that one can only apply into the workshop of their study lines at the university, which is not the case (as with comm&media). It could be rephrased like _"Kérjük, jelentkezését csak olyan műhelyekbe adja be, amelyek munkájában komolyan részt tud venni. A műhelyek egymástól függetlenül dönthetnek a meghallgatásáról."_

tests/Feature/AdmissionTest.php
```diff
@@ -0,0 +1,316 @@
...
+        $this->createApplicant("applicant2", false, [$aurelion, $info]);
+        $this->createApplicant("applicant3", true, [$aurelion, $info]);
+
+
+        $this->actingAs($this->admin);
+        $response = $this->get(route('admission.applicants.index') . "?show_not_submitted=true");
+        $response->assertSee('applicant1');
+        $response->assertSee('applicant2');
+        $response = $this->get(route('admission.applicants.index'));
+        $response->assertSee('applicant3');
```
viktorcsimma, line 114: Could we test filtering on workshops too?

tests/Feature/AdmissionTest.php
```diff
@@ -0,0 +1,316 @@
...
+        $aurelion = Workshop::firstWhere('name', Workshop::AURELION);
+        $info = Workshop::firstWhere('name', Workshop::INFORMATIKA);
+        $angol = Workshop::firstWhere('name', Workshop::ANGOL);
+
+        $user->addRole(Role::get(Role::AGGREGATED_APPLICATION_COMMITTEE_MEMBER));
+
+        $this->createApplicant("applicant1", false, [$info->id]);
+        $this->createApplicant("applicant2", true, [$aurelion->id, $info->id]);
+        $this->createApplicant("applicant3", true, [$info->id, $angol->id]);
+        $this->createApplicant("applicant4", true, [$angol->id]);
```
viktorcsimma, line 196: Could we put this into a separate private function? It is used quite often.

viktorcsimma (changes requested):
Nice:) there seem to be some small things, but it is near production now.

After these changes, I think I'll merge this and #572 into #560; it's better to review them together.

One more thing: why did we remove so many fields from the personal information tab? They were useful.

kdmnk:
> One more thing: why did we remove so many fields from the personal information tab? They were useful.

What do you mean? @viktorcsimma

resources/views/auth/admission/index.blade.php
kdmnk: This is only to show "delete filter" button which only makes sense if a filter has been applied. An admin can see all applications, yes.

resources/views/auth/application/application.blade.php
kdmnk: I made it uppercase

resources/views/auth/admission/index.blade.php
viktorcsimma: Oh, I see; sorry then.

resources/views/auth/application/status_update_component.blade.php
kdmnk: We have 3 states: nothing, called in, admitted. It's hard to make it even with radio buttons. I agree that technically we can make someone admitted but not called in, which might be strange, but the system will eventually handle it as admitted. I think we can live with that, or make it nicer later.

resources/views/auth/application/questions.blade.php
kdmnk: "amelyek munkájában szakmailag részt tud venni"? Otherwise, agreed.

viktorcsimma: Nice:)

resources/views/auth/application/status_update_component.blade.php
viktorcsimma: OK; let's do it then later. Maybe we could open an issue for it.

viktorcsimma:
> > One more thing: why did we remove so many fields from the personal information tab? They were useful.
> 
> What do you mean? @viktorcsimma

It looks like this for me:

grafik

tests/Feature/AdmissionTest.php
kdmnk: Yes, but every time it's a little different. For tests, I think this is easier to debug. I already have a createApplicant helper :D

viktorcsimma: OK then:)

kdmnk:
Ah, nice catch. It's because those fields only appear for collegists, but the new applicants do not get that role. I will check for tenant role instead.

viktorcsimma (approved):
Looks good to me:) Let's merge this into `refactor_applications`.

kdmnk:
Or rather merge that first and change the PR's base here so we keep the PR's separated

PR #575, opened 2024-07-19
Make guest / collegist registration clearer

kdmnk:
On small displays, the register as guest option was not displayed on welcome page. Now both options are added in a dropdown.

(Based on the refactor_applications branch, merge that first)


viktorcsimma:
Nice; thank you:) But I think we could merge this directly into development. I'll open a new PR for that.

PR #576, opened 2024-07-19
Make guest / collegist registration clearer

viktorcsimma:
A variant of #575 that is based on development.

### DIFF
--- a/resources/views/welcome.blade.php
+++ b/resources/views/welcome.blade.php
@@ -26,6 +26,10 @@
     <header>
         <div class="navbar-fixed">
             <nav class="top-nav">
+                <ul id="register-dropdown" class="dropdown-content">
+                    <li><a href="{{ route('register') }}">@lang('general.register_collegist')</a></li>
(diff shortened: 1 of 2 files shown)


kdmnk:
Nice, I was tired to mess with git now. But who approves then? :D

resources/views/welcome.blade.php
```diff
@@ -78,13 +81,12 @@
...
-                $('.tooltipped').tooltip();
-            });
-        </script>
-    @endif
+    <script>
+        $(document).ready(function(){
+            $('.tooltipped').tooltip();
+            $(".dropdown-trigger").dropdown();
+        });
+    </script>
```
BertalanD, line 89: This is broken; it complains about `$` being undefined (it's live on staging, see there)

Can this be done without jQuery? Also, feels odd to have a dropdown for this on desktop.

cc @kdmnk

BertalanD: mine

PR #577, opened 2024-07-20
Last-minute fixes to application process

viktorcsimma:
Replaying still actual fixes of #572 onto new code, as well as of some newly spotted bugs.

- Newly registered applicants were redirected to the tenant verification page after registration. This is now fixed.
- There had been an error while creating WiFi access if the Neptun code was not unique.
- Changed strange order of operations in `deleteFile()` in AdmissionController.
- Renamed `update` route to `update_note` (that is more descriptive).
- An applicant now cannot modify the internal note of themselves, even through a well-formed POST request.
- Handling abnormally tall and slim profile pictures.
- Refactoring information box on top of application page.
- Requiring a number ~~between 1 and 5~~ for averages.

Those with "view" permission can still theoretically modify applicant data through a POST request. But we can [leave this for now](https://github.com/EotvosCollegium/mars/pull/572/files#r1685508000).

### DIFF
--- a/app/Utils/ApplicationHandler.php
+++ b/app/Utils/ApplicationHandler.php
@@ -20,8 +20,9 @@ trait ApplicationHandler
     {
         $data = $request->validate([
             'status' => 'required|in:extern,resident',
-            'graduation_average' => 'required|numeric',
-            'semester_average' => 'nullable',
+            'graduation_average' => 'required|numeric|min:1|max:5',
+            'semester_average' => 'nullable|array',
+            'semester_average.*' => 'numeric|min:1|max:5',
             'competition' => 'nullable',
             'publication' => 'nullable',
             'foreign_studies' => 'nullable',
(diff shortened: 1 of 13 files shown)


app/Utils/ApplicationHandler.php
```diff
@@ -20,8 +20,9 @@ public function storeQuestionsData(Request $request, User $user): void
     {
         $data = $request->validate([
             'status' => 'required|in:extern,resident',
-            'graduation_average' => 'required|numeric',
-            'semester_average' => 'nullable',
+            'graduation_average' => 'required|numeric|min:1|max:5',
+            'semester_average' => 'nullable|array',
+            'semester_average.*' => 'numeric|min:1|max:5',
```
kdmnk: The thing here is that if its from another country, it might not fall between 1-5.

kdmnk (approved):
LGTM with one comment that I let you decide.

app/Utils/ApplicationHandler.php
viktorcsimma: Oh, I see... I'll revert that then.

PR #580, opened 2024-07-22
Use the main navbar layout on the welcome page

BertalanD:
This automatically turns the top nav links ("Login", "Application for
college students" and "Register for tenants") automatically turn into a
sidenav on mobile.

The easiest way I could make the logo stay centered on desktop is to
make the side nav only show on mobile layouts. I made the navbar blade
take a parameter indicating if we want it fixed (like on the main UI).

The layout of other pages are unaffected.

This partially reverts #576 

Desktop layout:
image

Mobile layout:
image
image


viktorcsimma (approved):
Nice:) I've tested on staging and mobile too, it works. Thank you!

PR #581, opened 2024-07-22
filter on statuses on admission index page

kdmnk:


resources/views/auth/application/status_update_component.blade.php
```diff
@@ -1,7 +1,7 @@
 <div style="display: flex;flex-direction: column;align-items: center;justify-content: center;">
     <label>
         <input type="checkbox"
-               wire:click="callIn('{{$workshop->workshop_id}}')" @checked($workshop->called_in) />
+               wire:click="callIn('{{$workshop->workshop_id}}')" @checked($workshop->called_in || $workshop->admitted) />
```
viktorcsimma: It's nice:) the only problem is that if someone's been admitted and I click the "called in" checkbox, it becomes empty until reloading the page. But this is not urgent for now.

Or what about disabling the checkbox if the second one is checked?

viktorcsimma (approved):
Nice; thank you:) I have a small issue, but it's not urgent; merge as you wish.

resources/views/auth/application/status_update_component.blade.php
kdmnk: Thats a good idea

PR #584, opened 2024-07-22
Making tests pass (adding missing documentation comments, changing a hash algorithm)

viktorcsimma:
- Changed the algorithm choosing file names from MD5 to SHA-256. (Maybe it is not that critical as Deepsource has said, but still.)
- Added missing documentation comments.

I have one dilemma: at PrintController.php:299, it complains for a TODO. I don't feel myself ready to do that test, but it would probably not be wise to remove the comment. (And PrintControllerTest.php has even more TODOs...)

### DIFF
--- a/app/Http/Controllers/Dormitory/PrintController.php
+++ b/app/Http/Controllers/Dormitory/PrintController.php
@@ -33,6 +33,9 @@ class PrintController extends Controller
         $this->middleware('can:use,App\Models\PrintAccount');
     }
 
+    /**
+     * Returns the printer page.
+     */
     public function index()
     {
         return view('dormitory.print.app', [
@@ -41,6 +44,9 @@ class PrintController extends Controller
             ]);
     }
 
+    /**
+     * Handles when a user clicks the 'no paper' button.
+     */
     public function noPaper()
     {
         $reporterName = user()->name;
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/Dormitory/PrintController.php
```diff
@@ -303,18 +356,27 @@ private function updateCompletedPrintingJobs()
...
+     * Stores the uploaded file on the server
+     * with a hash as the filename,
+     * then returns the path.
+     */
     private function storeFile($file)
     {
         $path = $file->storePubliclyAs(
             '',
-            md5(rand(0, 100000) . date('c')) . '.pdf',
+            hash('sha256', rand(0, 100000) . date('c')) . '.pdf',
```
horcsinbalint, line 368: What is the difference between using md5 and sha256 in this context?

horcsinbalint:
Why does this have high priority? I cannot see any kind of plausible security vulnerability that can leverage the fact that we use md5 instead of sha256 in that context (the only thing an attacker is able to control is the date). Increasing the maximum for the random generation, or simply base64 encoding 48 (or some other amount) random bytes and using that as a string could be better.

viktorcsimma:
Well, the main reason is that it annoys DeepSource:D I can imagine some kind of attack where someone deliberately uploads a file that gets the same hash as another, thereby overwriting it. But I don't insist; I can revert that if you say so.

horcsinbalint:
I meant that in that given context I don't think it really matters whether you are using md5 or sha256. Why don't we ditch rolling our own crypto and use something like the following?
```php
$path = $file->storePublicly(
    '',
    'printing'
);
```

viktorcsimma:
Now, #598 is simpler. I'll close this; maybe the documentation comments could be added later.

PR #585, opened 2024-07-23
Always prefer Hungarian if the user speaks it

BertalanD:
Even if the user has English as the highest priority entry in their Accept-Language header, we should select Hungarian by default if it's one of the languages that the user speaks.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -68,9 +68,15 @@ Route::get('/privacy_policy', [HomeController::class, 'privacyPolicy'])->name('p
 Route::get('/img/{filename}', [HomeController::class, 'getPicture']);
 Route::get('/setlocale/{locale}', [HomeController::class, 'setLocale'])->name('setlocale');
 
-Auth::routes(); //check \Laravel\Ui\AuthRouteMethods
+Auth::routes(['register' => false]); //check \Laravel\Ui\AuthRouteMethods
 
-Route::get('/register/guest', [RegisterController::class, 'showTenantRegistrationForm'])->name('register.guest');
+/* Override the /register route from Auth::routes to force Hungarian */
+Route::prefix('/register')->group(function () {
+    Route::post('/', [RegisterController::class, 'register']);
+
+    Route::get('/', [RegisterController::class, 'showRegistrationForm'])->name('register')->middleware(OnlyHungarian::class);
+    Route::get('/guest', [RegisterController::class, 'showTenantRegistrationForm'])->name('register.guest');
+});
 
 Route::middleware([Authenticate::class, LogRequests::class])->group(function () {
     /** Routes that needs to be accessed during the application process */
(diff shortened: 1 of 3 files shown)


BertalanD:
This should prevent the site appearing in English for most applicants (well, if they configured their device to also pass `hu` in Accept-Language).

kdmnk:
The application page should be in hungarian only (see OnlyHungarian Middleware). Otherwise, LGTM, but please add some comments to describe the different variables.

kdmnk (approved):
See comment above

BertalanD:
> The application page should be in hungarian only (see OnlyHungarian Middleware)

Good point. I'll add that to this PR.

BertalanD:
It seems that that's already the case for the application form, it's just the initial registration form that might show up in English.

routes/web.php
```diff
@@ -68,9 +68,15 @@
...
+Auth::routes(['register' => false]); //check \Laravel\Ui\AuthRouteMethods
 
-Route::get('/register/guest', [RegisterController::class, 'showTenantRegistrationForm'])->name('register.guest');
+/* Override the /register route from Auth::routes to force Hungarian */
+Route::prefix('/register')->group(function () {
+    Route::post('/', [RegisterController::class, 'register']);
+
+    Route::get('/', [RegisterController::class, 'showRegistrationForm'])->name('register')->middleware(OnlyHungarian::class);
+    Route::get('/guest', [RegisterController::class, 'showTenantRegistrationForm'])->name('register.guest');
+});
```
kdmnk, line 79: I would not overcomplicate things, if the Locale Middleware also affects the register page it should be enough. Maybe in the registration page put the description explicitly in hungarian and english as well (so also specify 'hu').
I'm also fine with the current state but feels too much for a simple and built-in feature.

kdmnk (approved):
I approve, but take a look at the comment.
Also, will this allow Hungarian users to chose the site to show another language if they want to? If not, we loose the whole point of a language selector.
I'm still confused what getLocales and getPreferredLanguage returns.

routes/web.php
BertalanD: I do not understand what you're suggesting. How else would I make `OnlyHungarian` affect `/register`?

kdmnk: I was referring to app/Http/Middleware/Locale.php

BertalanD:
> Also, will this allow Hungarian users to chose the site to show another language if they want to? If not, we loose the whole point of a language selector.

Yes, functionality regarding the language selector is unaffected. This only matters for deciding what language to show when the `locale` cookie is not set -- meaning language selector hasn't been used  (i.e. the second argument to `cookie()` specifies the default value).

> I'm still confused what getLocales and getPreferredLanguage returns.
- `getLocales` returns  an array of *all* language codes (e.g. `['en-US', 'en', 'hu']`) from the user's Accept-Language header (which is something like `en-US,en;q=0.9,hu;q=0.8` for me)
- `getPreferredLanguage` returns the highest priority *accepted* language which is present in the *available* languages (passed via the parameter). For instance, if the user agent passes `zh;q=0.9,de;q=0.8,hu;q=0.7`, it will select the German translation.

PR #589, opened 2024-07-23
Changing status filter to radio buttons

viktorcsimma:
As in the title. The filter now looks like this; I think this one is more understandable and also passes the tests:

grafik

Oh... ouch; now I see you have progressed since then. Not urgent now; let me know what you think.


horcsinbalint:
Superseeded by #590 as requested by the Student Council

PR #590, opened 2024-07-23
Improve applicant filtering

BertalanD:
- Made status filters exclusive.
- Users that only have access to one workshop can also filter by status.


horcsinbalint (approved):
LGTM!

viktorcsimma:
Wait! That's something like I've just done in #589

PR #595, opened 2024-07-24
Fix special chars in `x-input.text` being doubly escaped

BertalanD:
Per the [Laravel docs](https://laravel.com/docs/11.x/blade#passing-data-to-components):

> PHP expressions and variables should be passed to the component via attributes that use the : character as a prefix

Otherwise, HTML special characters get entity-escaped both when being passed to `x-input.text` and inside that component to the HTML `text` element. Without this, entering `a < b` would render as `a &lt; b`, then after an edit: `a &amp;lt; b`, etc...

Obsoletes #456

### DIFF
--- a/resources/views/components/input/text.blade.php
+++ b/resources/views/components/input/text.blade.php
@@ -4,7 +4,7 @@
     <input
         id="{{$id}}"
         class="validate @error($id) invalid @enderror"
-        value="{{old($id) ?? $attributes->get('value')}}"
+        value="{{old($id, $value ?? '')}}"
         {{-- Default values + other provided attributes --}}
         {{$attributes->whereDoesntStartWith('value')->merge([
             'type' => 'text',
--- a/app/View/Components/Input/Text.php
+++ b/app/View/Components/Input/Text.php
@@ -13,13 +13,15 @@ class Text extends Input
      * Create a new text input instance.
      *
      * @param  string  $helper  helper message
+     * @param  ?string $value   pre-filled contents
      * @return void
      */
-    public function __construct($id, $withoutLabel = false, $text = null, $s = 12, $m = null, $l = null, $xl = null, $onlyInput = false, $helper = null)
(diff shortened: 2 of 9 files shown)


resources/views/components/input/text.blade.php
```diff
@@ -4,7 +4,7 @@
     <input
         id="{{$id}}"
         class="validate @error($id) invalid @enderror"
-        value="{{old($id) ?? $attributes->get('value')}}"
```
kdmnk, line 7: The old value is needed when the controller redirects back with validation errors, it keeps the previously filled values. Please check if you still have that behavior.

kdmnk (changes requested):
Also keep it consistent with all our input components (checkboxes etc)

resources/views/components/input/text.blade.php
BertalanD: It seems okay to me, but not sure if I'm testing the right behavior (I know for fact that some validation is done on the client side too).

How would you recommend testing this?

horcsinbalint: I think `old` is already handled in `app/View/Components/Input/Text.php:23`

viktorcsimma: I've now tested it by entering a non-numeric value to closed university semester grades (those are only checked by the validator). It worked for me too.

kdmnk: Ah, I see, sorry.

kdmnk (approved):
Thanks for the clarification, looks good then, however I'd still prefer to have the same behaviour among the components. It triggers me a bit that you handle the old value in the component php file while the other components have it in their blade files :D But I can live with that

viktorcsimma:
Also, @BertalanD could you resolve merge conflicts?

BertalanD:
> triggers me a bit that you handle the old value in the component php file while the other components have it in their blade files

It turns out that moving the call to `old()` back to the blade works too; the key part of the fix was taking the attribute as a parameter of the component's constructor instead of accessing it via `$attributes`.

Also note to future me: there is some weird caching behavior going on, and I had to run `php artisan cache:clear` `php artisan optimize:clear` between testing the different variants to have things work properly.

PR #596, opened 2024-07-24
Add hint for applicants with existing accounts to log in

BertalanD:
This is shown unconditionally on the applicant registration page, which is easier than trying to specifically handle duplicate email address errors. I don't like that we're showing 4 (!) such messages on this page, we might want to come up with a nicer solution in the future.

image


viktorcsimma (approved):
Good for now; sorry it has taken such a long time.

On the long run, we could maybe put this information in the error message when the email already exists. But that means we need to add a mechanism for this before validation.

PR #597, opened 2024-07-24
Making Neptun code nullable for applicants

viktorcsimma:
Applicants could not leave the Neptun code and the educational email address empty, although many of them are going to receive these much later than the beginning of the application period.

Now they can leave them empty, but have to fill them before final submission.


horcsinbalint:
I have a concern on what happens if someone else has already entered the same Neptun code maliciously. But it was already like this before and I think it is probably out of scope of this PR. So LGTM!

horcsinbalint (approved):

PR #598, opened 2024-07-24
Adapting tests to new filters (and making them pass)

viktorcsimma:
As in the title. Now, the tests do pass.


BertalanD (approved):
LGTM!

Maybe we could have a test for the 403 error, too? E.g. a regular application committee member trying to pass the `unsubmitted` or `everybody` filter (to ensure the validation doesn't *only* happen on the UI side and can't be bypassed by manually setting the query parameter)

viktorcsimma:
That's included:) see the assertDontSeeUnsubmitted auxiliary function in AdmissionTest.

BertalanD:
🤦‍♂️ doubly looks good, then!

PR #600, opened 2024-07-25
Deploying more urgent fixes to the application process

viktorcsimma:
As in the title. Would include #597, #598 and #599.


viktorcsimma:
I've changed my mind, I think I'll rather deploy them by hand.

PR #604, opened 2024-07-27
Displaying the status update panel for those with no role whatsoever

### LINKED ISSUE #603: Adding instructions for those with neither an application nor a tenant role
There have been some who have had an account but currently have no tenant status. For them, the instructions do not show up, because those are displayed only to those who have a tenant status but with a `tenant_until` field in the past.

There, we could display two options: one to request tenant status again and one which redirects to the application form. WDYT?

viktorcsimma:
Closes #603.

Displays the `tenant-update` panel (including the application link) for those having no roles at all (not even a tenant role), as is with the case for many prospective applicants.

### DIFF
--- a/app/Http/Controllers/Secretariat/UserController.php
+++ b/app/Http/Controllers/Secretariat/UserController.php
@@ -314,6 +314,7 @@ class UserController extends Controller
         $validator->validate();
 
         $date = min(Carbon::parse($request->tenant_until), Carbon::now()->addMonths(6));
+        $user->addRole(Role::get(Role::TENANT));
         $user->personalInformation->update(['tenant_until' => $date]);
         $user->internetAccess()->update(['has_internet_until' => $date]);
 
@@ -431,14 +432,15 @@ class UserController extends Controller
     }
 
     /**
-     * Updates a tenant to an applicant
+     * Updates a tenant (or a user with no status) to an applicant
      */
     public function tenantToApplicant()
     {
-        if (!user()->isTenant() || user()->isCollegist(alumni: false)) {
-            return abort(403);
+        if (user()->isCollegist(alumni: false)) {
+            return abort(403, 'You are already a collegist.');
         }
         $user = user();
+        $user->update(['verified' => false]);
         $user->personalInformation()->update(['tenant_until' => null]);
         $user->removeRole(Role::get(Role::TENANT));
         $user->application()->create();
--- a/app/Http/Middleware/RedirectToStatusUpdate.php
+++ b/app/Http/Middleware/RedirectToStatusUpdate.php
@@ -38,6 +38,10 @@ class RedirectToStatusUpdate
         if ($user->application) {
             return $next($request);
         }
+        // Ignore those having any role
+        if ($user->roles()->exists()) {
+            return $next($request);
+        }
         return redirect(route('users.tenant-update.show'));
     }
 }
--- a/app/Http/Middleware/RedirectToStatusUpdate.php
+++ b/app/Http/Middleware/RedirectToStatusUpdate.php
@@ -5,11 +5,11 @@ namespace App\Http\Middleware;
 use Closure;
 use Illuminate\Http\Request;
 
-class RedirectTenantsToUpdate
+class RedirectToStatusUpdate
 {
     /**
      * Redirects tenants to update their tenant_until property if it is in the past.
-     * This way we can distinguish between the active and inactive tenants.
+     * Also appears to users with no status whatsoever (neither collegist nor tenant).
      *
      * @param \Illuminate\Http\Request $request
      * @param \Closure(\Illuminate\Http\Request): (\Illuminate\Http\Response|\Illuminate\Http\RedirectResponse) $next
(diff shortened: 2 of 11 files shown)


app/Http/Controllers/Secretariat/UserController.php
```diff
@@ -314,6 +314,7 @@ public function updateTenantUntil(Request $request, User $user)
         $validator->validate();
 
         $date = min(Carbon::parse($request->tenant_until), Carbon::now()->addMonths(6));
+        $user->addRole(Role::get(Role::TENANT));
```
kdmnk, line 317: this might add multiple Tenant roles for the same user (tenants have to extend their status regularly)

app/Http/Controllers/Secretariat/UserController.php
```diff
@@ -431,14 +432,15 @@ public function showTenantUpdate()
...
     /**
-     * Updates a tenant to an applicant
+     * Updates a tenant (or a user with no status) to an applicant
      */
     public function tenantToApplicant()
     {
-        if (!user()->isTenant() || user()->isCollegist(alumni: false)) {
-            return abort(403);
+        if (user()->isCollegist(alumni: false)) {
+            return abort(403, 'You are already a collegist.');
```
kdmnk, line 440: localize error messages

app/Http/Controllers/Secretariat/UserController.php
```diff
@@ -431,14 +432,15 @@ public function showTenantUpdate()
...
         }
         $user = user();
+        $user->update(['verified' => false]);
```
kdmnk, line 443: applicants might also be verified (see reapplication for resident status). Are you sure you need this?

app/Http/Middleware/RedirectToStatusUpdate.php
```diff
@@ -34,6 +34,14 @@ public function handle(Request $request, Closure $next)
         if (!$user->needsUpdateTenantUntil()) {
             return $next($request);
         }
+        // Ignore those having an application in progress
+        if ($user->application) {
```
kdmnk, line 38: I don't like that the existence of a php value is used to differentiate user types. First, I don't trust PHP, secondly anyone can just create an application as it is created automatically if you visit the application page. It might not make a difference here but we should aim for a clearer approach.

app/Models/User.php
```diff
@@ -834,11 +834,12 @@ public function isCurrentTenant(): bool
...
     /**
      * @return bool if the user needs to update their tenant status
-     * A user needs to update their tenant status if they are a tenant and their tenant_until date is in the past.
+     * A user needs to update their tenant status if they are a tenant and their tenant_until date is in the past;
+     * or if they have no status whatsoever.
      */
     public function needsUpdateTenantUntil(): bool
     {
-        return $this->isTenant() && !$this->isCurrentTenant();
+        return !$this->isCollegist() && !$this->isCurrentTenant();
```
kdmnk, line 842: Won't this tigger for teachers and so on?

tests/Feature/AdminInternetControllerTest.php
```diff
@@ -23,6 +23,8 @@ public function test_no_access_admin_internet_page(): void
     {
         $response = $this->actingAs($this->user)->get(route('internet.admin.index'));
 
+        echo substr(json_encode($response), 0, 500);
```
kdmnk, line 26: delete

routes/web.php
```diff
@@ -76,6 +76,8 @@
 
     Route::get('/', [RegisterController::class, 'showRegistrationForm'])->name('register')->middleware(OnlyHungarian::class);
     Route::get('/guest', [RegisterController::class, 'showTenantRegistrationForm'])->name('register.guest');
+
+    Route::post('/rerequest_tenant_status', [RegisterController::class, 'rerequestTenantStatus'])->name('register.rerequest_tenant_status');
```
kdmnk, line 80: undefined function

kdmnk (changes requested):

app/Http/Controllers/Secretariat/UserController.php
viktorcsimma: I see... I put it into an `if` then.

app/Models/User.php
viktorcsimma: That is filtered at RedirectToStatusUpdate.php:42; i.e. anyone who has any role is discarded. But you are right we should move that logic here.

kdmnk (changes requested):
I don't like this approach, we are making edge cases and adding exceptional behaviour (see test cases) that will not make our life easier. 

We should make the system in a way that we don't have (approved) users without any role or even if we have, we can ignore them by all means.

Currently, I can identify the following user types:
- not approved applicant, without any role (which is fine, they can access the application page)
- approved applicant with collegist role (reapplying collegist)
- approved applicant with (older) tenant role
- other users, all with some role. 

And for the approved collegists and tenants, we need to offer the ability to start the application process. 

It works in my head (we should document this btw) but idk then what goes wrong.

Ps. with this above, the tenantToApplicant should do the following:
Remove their role and make them non-verified.
For reapplying collegists, technically we should not do anything, they can make an application.

routes/web.php
viktorcsimma: Yepp; I've forgotten about that. It was a remnant of an earlier approach.

viktorcsimma:
Now you're talking; I agree.

Mhm... so for now, could we solve this simply by making all "roleless" users unverified? That would be a single SQL command.

The problem is probably that they have somehow lost their tenant role; probably when they reached `tenant_until`. We should investigate that.

Anyway, I'll close this for now, if you agree.

PR #608, opened 2024-07-29
Letting tenants view rooms

viktorcsimma:
As in the title. Based on a user request.

### DIFF
--- a/app/Policies/RoomPolicy.php
+++ b/app/Policies/RoomPolicy.php
@@ -24,6 +24,7 @@ class RoomPolicy
             Role::SECRETARY,
             Role::STAFF,
             Role::COLLEGIST,
+            Role::TENANT,
             Role::SYS_ADMIN
         ]);
     }


kdmnk:
I don't know if the collegists can see tenants now, but I'm not sure if the people want that anyone in the building can see where do they live. Thoughts?

viktorcsimma:
As far as I can see, neither collegists see tenants nor vice versa.

Actually, it's a wish of some former collegists who want to interact with people inside. Maybe we could tie this to the alumni role. WDYT?

kdmnk:
> As far as I can see, neither collegists see tenants nor vice versa.

Then what's the point of this issue?
btw, I just wanted to raise my concerns, I don't really care or want to make a statement. Ask the student council what makes sense.

The code itself is fine, I'll approve it

kdmnk (approved):

viktorcsimma:
It's been voted down; I close this. Probably we should solve these user requests individually.

PR #612, opened 2024-08-02
Add notice about applying with missing Neptun code

BertalanD:
We were asked to make applying possible without a Neptun code in the last 3 days of the application period. This is a worst-case scenario and will require code changes; let's ask the applicants to notify us if they haven't received the code by then.

### DIFF
--- a/resources/views/auth/application/app.blade.php
+++ b/resources/views/auth/application/app.blade.php
@@ -40,6 +40,12 @@
                             véglegesítse jelentkezését a lap alján lévő gombra kattintva.
                             Kérjük, figyeljen a határidőre, mert utána már nem lesz lehetősége véglegesítésre.</li>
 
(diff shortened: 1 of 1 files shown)


kdmnk:
Isn't it easier to just make it nullable for everyone? I feel like most of them will insert it anyways and we can update or ask them to update later anytime.

kdmnk (approved):

BertalanD:
I would prefer not to make it nullable unless absolutely necessary, since we need it for generating the Wi-Fi credentials (are there other places where we assume that a Neptun code exists and uniquely identifies a user?)

kdmnk:
There is a fallback value for wifi credentials (the same for guests). Other than that we don’t really use the neptun code anywhere AFAIK.

BertalanD:
Also approved by the Secretariat. Merging.

PR #613, opened 2024-08-04
Make university e-mails optional for applicants

BertalanD:
They are expected to be set up for first-year students even later than Neptun codes (Wednesday being the earliest date), so we cannot have it be mandatory.


horcsinbalint (approved):
LGTM

PR #615, opened 2024-08-05
Make registration less misleading

horcsinbalint:
## Description
Make registration easier to understand

## Changes
- Added information on tenant page
- Updated language for these pages


resources/views/auth/register.blade.php
```diff
@@ -15,15 +15,22 @@
...
                     <a href="{{route('register.guest')}}" style="text-decoration: underline">@lang('registration.information_tenant', [], 'en')</a>
                 </blockquote>
+                <blockquote>
+                    <a href="{{route('register.guest')}}" style="text-decoration: underline">@lang('registration.information_tenant', [], 'hu')</a>
+                </blockquote>
                 <blockquote>
                     Ha már rendelkezel {{ config('app.name') }} fiókkal, <a href="{{ route('login') }}">belépést</a> követően adhatod le jelentkezésed.
                 </blockquote>
+                @else
+                <blockquote>
```
BertalanD: Add `lang="hu"` attribute here.

BertalanD (approved):
We could bikeshed a lot on the exact wording, but I think this serves the purpose of clarifying the two forms of registration well.

We should revert this (and the other notices) after admissions are done and figure out a more elegant solution, though.

viktorcsimma:
Huh; thank you very much:)

PR #617, opened 2024-08-07
Clarify the expected format of CVs

BertalanD:
https://github.com/user-attachments/assets/6675e2b7-f285-4524-8cf8-1b08807c706a

### DIFF
--- a/resources/views/auth/application/files.blade.php
+++ b/resources/views/auth/application/files.blade.php
@@ -11,7 +11,12 @@
             <blockquote>
                 A pályázatnak az alábbiakat kell tartalmaznia:
                 <ul style="margin-left:20px;margin-top:0">
-                    <li style="list-style-type: circle !important">önéletrajz (hagyományos, leíró jellegű)</li>
+                    <li style="list-style-type: circle !important"><em>hagyományos, leíró jellegű</em> önéletrajz
+                        <i class="material-icons tooltipped" style="font-size: 1em; vertical-align: -0.2em; cursor: default"
+                           data-tooltip="teljes mondatokból álló, összefüggő szöveg (NEM Europass-jellegű)">
+                           info_outline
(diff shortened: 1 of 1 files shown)


viktorcsimma (changes requested):
Nice:) I have two suggestions for the text itself; WDYT? I do not insist on them.

resources/views/auth/application/files.blade.php
```diff
@@ -11,7 +11,12 @@
             <blockquote>
                 A pályázatnak az alábbiakat kell tartalmaznia:
                 <ul style="margin-left:20px;margin-top:0">
-                    <li style="list-style-type: circle !important">önéletrajz (hagyományos, leíró jellegű)</li>
+                    <li style="list-style-type: circle !important"><em>hagyományos, leíró jellegű</em> önéletrajz
+                        <i class="material-icons tooltipped" style="font-size: 1em; vertical-align: -0.2em; cursor: default"
+                           data-tooltip="Teljes mondatokból álló, összefüggő szöveg (NEM Europass-jellegű)">
```
viktorcsimma: My suggestion would be this:

`data-tooltip="teljes mondatokból álló, összefüggő szöveg (**nem** Europass-jellegű)">`

This way, it would pass more into the context (which does not use capital letters) and could look a bit more professional.

It would be even better to have `nem` with a bold font if that could be solved. But I can live without that.

BertalanD: `data-tooltip` only takes plain text, unfortunately (surrounding it with `*`s gives the impression of failed text formatting...)

Agree on the capitalization.

viktorcsimma: OK:) then let's stay with the capitalization.

viktorcsimma (approved):

PR #618, opened 2024-08-08
Finalise application process

### LINKED ISSUE #327: Make the secretary able to handle some changes to submitted applications
- Upload files
- Edit collegist/extern status
- Edit applied workshops

This should be enough

kdmnk:
This PR implements the missing finalise applications feature. 

It's on a new page now where the final names and statistics can be seen.
It is also possible to change who has been admitted as resident or extern.


app/Http/Controllers/Auth/AdmissionController.php
```diff
@@ -165,54 +168,82 @@ public function updateNote(Request $request, Application $application): Redirect
         return redirect()->back();
     }
 
+    /**
+     * Show the finalize page with final names and statistics
+     */
+    public function indexFinalize(): View
+    {
+        $this->authorize('finalize', Application::class);
+        $admitted = Application::query()->with(['user', 'applicationWorkshops'])->admitted()->get()->sortBy('user.name');
```
viktorcsimma: Isn't it more efficient to do the ordering before the `get()`, while still in the query? With an `orderBy`.

app/Http/Controllers/Auth/AdmissionController.php
```diff
@@ -165,54 +168,82 @@ public function updateNote(Request $request, Application $application): Redirect
...
+            // delete data for not admitted users
+            $files = File::query()
+                ->whereIn('application_id', $not_admitted_applications->pluck('id')) // application files
+                ->orWhereIn('user_id', $not_admitted_applications->pluck('user_id')); // profile pictures
+            foreach ($files->get() as $file) {
+                Storage::delete($file->path);
+            }
+            $files->delete();
+            // soft deletes application, keep them for future reference
+            // (see https://github.com/EotvosCollegium/mars/issues/332#issuecomment-2014058021)
```
viktorcsimma, line 237: Does the college qualify as a 'higher education institute'? (AFAIK, only the university itself has legal personality, not even the faculties.) But I'm obviously not a lawyer, and I can live without purging them anyway:)

app/Http/Controllers/Auth/AdmissionController.php
```diff
@@ -165,54 +168,82 @@ public function updateNote(Request $request, Application $application): Redirect
...
+                $application->user->setStatusFor($this->semester(), SemesterStatus::ACTIVE);
+                $application->user->internetAccess->extendInternetAccess($this->semester()->getStartDate()->addMonth());
+            }
+            // delete data for not admitted users
+            $files = File::query()
+                ->whereIn('application_id', $not_admitted_applications->pluck('id')) // application files
+                ->orWhereIn('user_id', $not_admitted_applications->pluck('user_id')); // profile pictures
+            foreach ($files->get() as $file) {
+                Storage::delete($file->path);
+            }
```
viktorcsimma: Are these also only soft-deleted? If applications are kept, the files belonging to them should be kept as well.

app/Http/Controllers/Auth/AdmissionController.php
```diff
@@ -165,54 +168,82 @@ public function updateNote(Request $request, Application $application): Redirect
...
+            $files->delete();
+            // soft deletes application, keep them for future reference
+            // (see https://github.com/EotvosCollegium/mars/issues/332#issuecomment-2014058021)
+            Application::whereIn('id', $admitted_applications->pluck('id'))->delete();
+            Application::whereNotIn('id', $admitted_applications->pluck('id'))->forceDelete();
+            ApplicationWorkshop::query()->delete();
+
+            // Note: users with a not submitted applications will also be deleted
```
viktorcsimma: Nitpick: there is an unnecessary `a` there.

app/Http/Controllers/Auth/AdmissionController.php
```diff
@@ -165,54 +168,82 @@ public function updateNote(Request $request, Application $application): Redirect
...
+        $not_admitted = Application::query()->whereNotIn('id', $admitted->pluck('id'))->get();
+        return view('auth.admission.finalize', [
+            'semester' => $this->semester(),
+            'admitted_applications' => $admitted,
+            'users_to_delete' => User::query()
+                ->withoutGlobalScope('verified')
+                ->with('application')
+                ->whereIn('id', $not_admitted->pluck('user_id'))
+                //ignore users with any existing role
+                ->whereDoesntHave('roles')
```
BertalanD: FYI: At the moment, even newly registered applicants get the printer role by default, so I don't think this will end up deleting anybody as-is. (last year's applicants are still present in the `users` table too)

https://github.com/EotvosCollegium/mars/blob/fabec6fb25c18d1c98a004e438094e62f5546fb2/app/Http/Controllers/Auth/RegisterController.php#L115

Probably easier to jump into the database and fix this manually, rather than changing the code here to handle this quirk.

app/Http/Controllers/Auth/AdmissionController.php
```diff
@@ -165,54 +168,82 @@ public function updateNote(Request $request, Application $application): Redirect
...
+                ->orderBy('name')
+                ->get()
```
BertalanD: Can we factor out this query and the identical one in `finalize` into a helper function?

Also, this looks prone to race conditions; what happens if someone messes with admission statuses between this page is rendered and the finalize button is pressed?

app/Http/Controllers/Auth/AdmissionController.php
BertalanD, line 237: @horcsinbalint

app/Http/Controllers/Auth/AdmissionController.php
kdmnk: Huh, true... But this won't be the case after https://github.com/EotvosCollegium/mars/pull/394 has been merged. 
But even before that we can just delete that role.

app/Http/Controllers/Auth/AdmissionController.php
kdmnk: Good point. But in real life I expect that the workshop committees will have a deadline to finalise the statuses. Or everything will be handled by the secretary/sys admins. 
Let's just pay attention to this as I don't have an easy solution now.

resources/views/auth/admission/index.blade.php
```diff
@@ -73,27 +73,9 @@
     <hr>
     <h6>Összesen: <b class="right">{{$applications->count()}} jelentkező</b></h6>
     @can('finalize', \App\Models\Application::class)
-        @if($applicationDeadline?->addWeeks(1) < now())
```
viktorcsimma, line 76: index.blade.php:53 adds a week to $applicationDeadline without making a copy of it. It works for now because it isn't used at the end, but maybe it would be more future-proof.

app/Http/Controllers/Auth/AdmissionController.php
BertalanD: > Let's just pay attention to this as I don't have an easy solution now.

Sounds good to me. But can you still do something with the code duplication for the query?

resources/views/auth/admission/finalize.blade.php
```diff
@@ -0,0 +1,80 @@
...
+    <a href="#!" class="breadcrumb">Felvételi véglegesítés</a>
+@endsection
+
+@section('content')
+    <div class="card" style="margin-top:20px">
+        <div class="card-content">
+            <div class="row">
+                <div class="col">
+                    Hogyha a felvételi eljárás befejeződött, akkor a felvett jelentkezőket itt lehet
+                    jóváhagyni. A felvettek státusza a {{ $semester->tag }} szemeszterre automatikusan aktív lesz.
```
viktorcsimma, line 14: This throws an exception if the periodic event is not yet set.

Also, for me, this is the spring semester and not the autumn one as it should be.

resources/views/auth/admission/finalize.blade.php
```diff
@@ -0,0 +1,80 @@
...
+                    @endforeach
+                </div>
+                <div class="col">
+                    <blockquote>A művelet végrehajtása előtt ajánlott biztonsági mentést készíteni az adatbázisról.</blockquote>
+                    <form id="finalize-application-process" method="POST"
+                          action="{{route('admission.finalize')}}">
+                        @csrf
+                        <x-input.button class="red" text="A felvételi lezárása"/>
+                    </form>
+                </div>
```
viktorcsimma, line 72: This should not appear if we have not reached `$applicationDeadline->copy()->addWeek()` yet.

app/Http/Controllers/Auth/AdmissionController.php
horcsinbalint: I have not checked thoroughly, but this part LGTM. If it deletes all the data for rejected users but keeps the data for admitted users, it should be fine AFAIK (I think it is the behavior that we should aim to achieve).

One question is what happens in cases where a student starts a `Jogorvoslati eljárás`. I would not want to go into details publicly about our backup strategy (as it is proprietary) but I think that should solve that issue.

app/Http/Controllers/Auth/AdmissionController.php
```diff
@@ -165,54 +168,82 @@ public function updateNote(Request $request, Application $application): Redirect
...
-        //            $usersToDelete->forceDelete();
-        //
-        //            RoleUser::where('role_id', Role::get(Role::APPLICATION_COMMITTEE_MEMBER)->id)->delete();
-        //            RoleUser::where('role_id', Role::get(Role::AGGREGATED_APPLICATION_COMMITTEE_MEMBER)->id)->delete();
-        //        });
-        //
-        //        Cache::clear();
-        // return back()->with('message', 'Sikeresen jóváhagyta az elfogadott jelentkezőket');
-        return back()->with('error', 'Még nincs implementálva.');
+        $this->authorize('finalize', Application::class);
```
viktorcsimma, line 207: Here, we should check whether we have reached `$applicationDeadline->copy()->addWeek()`.

Or should we abolish this requirement? (I think it is useful, but I do not insist on it.) If not more, I think we should not allow finalisation while the application period is open. (The finalisation page could be shown, that is useful; but without the option to actually finalise applications.)

app/Http/Controllers/Auth/AdmissionController.php
horcsinbalint: But the question about `Jogorvoslati eljárás` is not a real issue IMO.

viktorcsimma (changes requested):
Also, a seeded applicant did not get workshops nor educational information; is this specific to seeded users or a general problem? I think at least workshops should not be a problem.

app/Http/Controllers/Auth/AdmissionController.php
viktorcsimma: I agree; we don't need the printer role. No one can print without paying for admins before, anyway.

app/Http/Controllers/Auth/AdmissionController.php
horcsinbalint: The College itself is not a `felsőoktatási intézmény`, but it is rather a `szervezeti egység` of the university which is a `felsőoktatási intézmény`. So I think the College should comply with the `Nemzeti Felsőoktatási Törvény`. Where the GDPR is concerned the `adatkezelő` is the university and the `felelős szervezeti egység` is the College AFAIK.

app/Livewire/ApplicationRoleStatusUpdate.php
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Update whether the applicant has been admitted as resident or not.
+     * @param $workshop
+     */
+    public function switchResidentRole()
+    {
+        $this->application->update(['admitted_for_resident_status' => !$this->application->admitted_for_resident_status]);
+        $this->lastUpdated = Carbon::now();
+    }
```
BertalanD: Does this need access control?

app/Http/Controllers/Auth/AdmissionController.php
viktorcsimma: I see... I'm not confident in these things; let this be so then.

app/Http/Controllers/Auth/AdmissionController.php
kdmnk: Then I would need to join the users table... If 'with' has other semantics (which I guess) then we still have some overhead. I think it is fine as it is

app/Http/Controllers/Auth/AdmissionController.php
kdmnk: I only delete the not admitted applications' files. The others just stay there.

app/Http/Controllers/Auth/AdmissionController.php
kdmnk: I wanted to delete that part (maybe I left some parts of this logic). Originally, i just wanted to avoid accidental finalisations but now we have a separate page only for that, I think it is clear now.

BertalanD: FWIW, removing that sounds okay to me (the one-week delay struck me as arbitrary when I first saw it)

BertalanD:
@coderabbitai review

app/Http/Controllers/Auth/AdmissionController.php
viktorcsimma: OK:)

app/Http/Controllers/Auth/AdmissionController.php
viktorcsimma: Oh, sorry; I've missed that.

app/Http/Controllers/Auth/AdmissionController.php
viktorcsimma: I think it would still be useful to prevent finalisation while the application period is open. But I can live without that; just remove the remainders and warnings related to that logic then.

resources/views/auth/admission/finalize.blade.php
kdmnk: The semester depends on the periodic event. You should set it for the autumn one. In the live system it looks fine.

app/Http/Controllers/Auth/AdmissionController.php
kdmnk: I have put it back but it only checks for the application deadline, does not add a week.

viktorcsimma: Nice:)

resources/views/auth/admission/finalize.blade.php
viktorcsimma: Oh, I see. Sorry then.

resources/views/auth/admission/index.blade.php
```diff
@@ -73,26 +72,10 @@
...
-                                felvételihez kapcsolódó (felvételiztető) jog elvételre kerül.
-                                A "Véglegesítve" és "Behívva" státuszú jelentkezőket előbb el kell utasítani, vagy fel
-                                kell venni.
-                            </div>
-                            <x-input.button class="red right" text="Felvételi lezárása"/>
-                        </form>
-                    </div>
-                </div>
-            </div>
+        @if($applicationDeadline && $applicationDeadline < now()))
```
viktorcsimma: Here, there is an extra `)` which appears on the generated page.

viktorcsimma (changes requested):
LGTM now:) one more small change and we can go ahead, I think.

viktorcsimma:
@coderabbitai review

resources/views/auth/admission/index.blade.php
kdmnk: I don't have my laptop for the rest of the weekend, if you want to merge could you fix it?

PR #622, opened 2024-08-09
Use applied workshop list for applicant export

BertalanD:
`$user->workshops` is only set for users that have already been admitted to the Collegium. Use the workshop list from application data for the Excel applicant export instead.

### DIFF
--- a/app/Exports/ApplicantsExport.php
+++ b/app/Exports/ApplicantsExport.php
@@ -81,7 +81,9 @@ class ApplicantsExport implements FromCollection, WithTitle, WithMapping, WithHe
                 return $studyLine->getNameWithYear();
             })->implode(', '),
             implode(",", $user->faculties->pluck('name')->toArray()),
-            implode(",", $user->workshops->pluck('name')->toArray()),
+            $application->applicationWorkshops->map(function ($workshop) {
+                return $workshop->workshop->name;
+            })->implode(","),
             $user->isResident() ? 'Bentlakó' : 'Bejáró',
             ($user->educationalInformation?->alfonso_language ?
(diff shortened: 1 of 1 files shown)


app/Exports/ApplicantsExport.php
```diff
@@ -81,7 +81,9 @@ public function map($application): array
                 return $studyLine->getNameWithYear();
             })->implode(', '),
             implode(",", $user->faculties->pluck('name')->toArray()),
-            implode(",", $user->workshops->pluck('name')->toArray()),
+            $application->applicationWorkshops->map(function ($workshop) {
```
viktorcsimma: Maybe it would be a bit more efficient to create a new BelongsToMany relationship between applications and workshops (where the connection table is the table of ApplicationWorkshops) and querying them – that way, it's only one request.

viktorcsimma (approved):
Only one thing, but I can live without it; merge as you wish.

app/Exports/ApplicantsExport.php
BertalanD: hmm, that does already exist (`appliedWorkshops`). why didn't I find it at first?

app/Exports/ApplicantsExport.php
```diff
@@ -81,9 +81,7 @@ public function map($application): array
                 return $studyLine->getNameWithYear();
             })->implode(', '),
             implode(",", $user->faculties->pluck('name')->toArray()),
-            $application->applicationWorkshops->map(function ($workshop) {
-                return $workshop->workshop->name;
-            })->implode(","),
+            $application->appliedWorkshops()->implode('name', ','),
```
viktorcsimma, line 84: Nice:) it's more elegant than what I've thought of.

PR #629, opened 2024-08-13
Alfonso refinements

viktorcsimma:
Closes #285. It cannot check whether someone is a language teacher, but can check exemptions for seniors and those who have been admitted during their masters' studies.

PR #631, opened 2024-08-19
Fixing a bug where the admittance semester has not yet been created

viktorcsimma:
#629 contained a bug which caused a crash when the model for the admittance semester has not yet been created.

This PR fixes that bug by assuming that the admittance semester is always an autumn semester, therefore not needing the semester model anymore.


BertalanD (approved):
LGTM! Thank you for fixing this.

PR #634, opened 2024-09-04
docker-dev-setup: Install node from nodesource.com

BertalanD:
This allows us to use native AArch64 builds on Apple Silicon Macs, so we no longer depend on Rosetta emulation working correctly (for some reason, it started to fail locally).


Trigary (approved):
LGTM

PR #643, opened 2024-09-14
Ensure that input IDs are unique on the checkout page

BertalanD:
Without this change, clicking an input field in the "Kiadás hozzáadása" section would trigger the respective field in the "Bevétel hozzáadása" section, as the IDs of the input fields were the same.

This commit adds prefixes to the IDs to indicate which form they belong to, while keeping the old "name" attributes to avoid having to change the backend code.


viktorcsimma (approved):
LGTM; thanks!

PR #646, opened 2024-09-29
Use periodic events for KKT/Netreg

### LINKED ISSUE #625: Use periodic events for kkt payments and internet deadline

kdmnk:
Some thing that was missing before.
I'm merging to https://github.com/EotvosCollegium/mars/pull/555, you can check both PRs.

When a KKT/Netreg period opens, now the collegists 'netreg_paid' will be set to false, and they have internet only until the payment deadline. If they pay, the netreg_paid=true will overwrite the deadline.

### DIFF
--- a/resources/views/network/admin/internet_access.blade.php
+++ b/resources/views/network/admin/internet_access.blade.php
@@ -1,8 +1,6 @@
 <span class="card-title">Internetelérés</span>
 <blockquote>
-    <p>Az aktuális internetelérés határidő: <span class="coli-text text-orange">{{ $activation_date }}</span></p>
-    <p>Collegisták esetén a Netreg befizetésekor automatikusan beállítja a rendszer a hozzáférésüket eddig a
-        dátumig.</p>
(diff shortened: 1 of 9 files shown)


resources/views/network/admin/internet_access.blade.php
```diff
@@ -134,13 +118,17 @@
                         minWidth: 150,
                     },
                     {
-                        title: "Internetelérés",
+                        title: "Netreg fizetett",
```
viktorcsimma, line 121: The `false` and `true` values could be localised; but as they are only visible for admins anyway, it's not crucial.

resources/views/network/admin/internet_access.blade.php
```diff
@@ -57,7 +41,7 @@
                 if (value) {
                     value = moment(value).format("YYYY. MM. DD. HH:mm");
                 } else {
-                    value = 'Nincs hozzáférés'
+                    value = 'N/A'
```
viktorcsimma, line 44: It could be something like 'Unlimited until next kkt period' for those having paid kkt; it would be easier to understand.

viktorcsimma (approved):
This seems to be a good idea:) take a look at these small things; then it's fine to merge into #555 and handle together with it.

PR #658, opened 2024-11-15
Remove the ability to mix free and paid pages

horcsinbalint:
## Description
There is a quite annoying bug where if a user prints using free pages, their free page pool is only decreased by p and not by n*p (where n is the number of copies, and p is the number of pages in the document). The quickest and easiest way of dealing with it is to only allow a print job to be fully paid or entirely free and deduct the printed pages correctly. The solution is not ideal but it is quite a rare occurrence where someone would like to use both free pages and pay for the remainder. (They can split it themselves.)

This bug is quite annoying as our registry shows only a fraction of the actual usage. I.e. ~ 5000 sheets was loaded into the printer during this semester, and our Mars instance shows only a few hundred printed pages. It would be crucial to understand how many pages were printed for different kinds of activities as it has already eaten roughly a quarter of our budget of the semester in half-a-semester.

## Changes

- Removed the ability to mix free and paid pages.
- Fixed the problem where multiple number of copies were only deducted once.

## Additional Notes
I am not refactoring the printer module for this PR.
To allow mixing these methods, the `PrintAccount` class would also need to be refactored but I do not think that it is worth the time to be invested as we will need additional refactoring as we are planning on introducing multiple printers.

### DIFF
--- a/app/Utils/Printer.php
+++ b/app/Utils/Printer.php
@@ -41,19 +41,29 @@ class Printer
 
         // If using free pages, check the amount that can be used
         if ($this->use_free_pages) {
+            $this->pages *= $this->number_of_copies;
+
             $this->calculateFreePagePool();
+
+            if ($this->pages > 0) {
+                return back()->withInput()->with('error', __('print.no_balance'));
+            }
+
+            // Print document
+            return $this->printDocument();
+        } else {
+
+            // Calculate cost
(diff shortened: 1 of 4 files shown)


horcsinbalint:
@coderabbitai full review

BertalanD (approved):

PR #667, opened 2024-12-22
Upgrading docker-dev-setup to Ubuntu 24.04

horcsinbalint:
Ubuntu 23.10 is EOL. For this reason docker-dev-setup no longer worked. Fixed by upgrading to 24.04 and PHP 8.3.


horcsinbalint:
Merging as it is only making the `docker-dev-setup` work again and should not have any impact on other parts of the software

PR #669, opened 2024-12-25
Fixes for print refactoring

horcsinbalint:
## Description

There are some issues in #394 and this PR's aim is to deal with some of these issues. This is still a draft, the issues addressed:

- Printing role is not deleted from the database if it comes from the previous versions
- Minor issue with executing data as program
- Minor issue with attributes
- Log results of commands

## Changes

- ~~Removed all references of Models in migrations, fixed the issues caused by this change in seeders~~ Moving to #670 except for printer role
- Drop printer role in the database
- Better check for IP addresses
- Fixed the problems with attributes
- Log results of commands

## Additional Notes

Old migrations have been changed multiple times which is not great practice. Referencing Models in migrations should be avoided as changing the Model might change older migrations. Migrations are usually the "way of propagating changes you make to your Models (adding a field, deleting a model, etc.) into your database schema" (https://docs.djangoproject.com/en/5.1/topics/migrations/). Therefore migrations should never ever reference the Models as their main purpose is to have a migration between two different versions of models. Once a migration is written, they should usually never be edited in the future (unless the language / framework changes the syntax). But functionally a migration should remain the same in its whole life regardless of how the application is developed.

~~Changing so many migrations should probably not be part of just a patch to another PR but that PR also had issues with it.~~ Moving to #670 (Namely older versions of the software created the printer role, and the patch in #394 no longer creates it in database/migrations/2019_10_13_130718_create_roles_table.php, but these roles stay in the database if a database was created with an earlier version.)

### DIFF
--- a/app/Utils/Process.php
+++ b/app/Utils/Process.php
@@ -7,6 +7,7 @@ use Symfony\Component\Process\Process as SymfonyProcess;
 
 class Process extends SymfonyProcess
 {
+
     /**
      * Run the process. If the app is in debug mode, the process will not be executed.
      *
@@ -14,10 +15,18 @@ class Process extends SymfonyProcess
      * @param array $env
      * @return int
      */
-    public function run(?callable $callback = null, array $env = []): int
+    public function run(?callable $callback = null, array $env = [], bool $log = true): int
     {
         if (config('app.debug') === false || config('commands.run_in_debug') === true) {
-            return parent::run($callback, $env);
+            $return_value = parent::run($callback, $env);
+            if($log) {
+                if($return_value === 0) {
+                    Log::info("Command: " . $this->getCommandLine() . " executed successfully.");
+                } else {
+                    Log::error("Command: " . $this->getCommandLine() . " failed with error code: " . $return_value . "\nWith output: " . $this->getOutput() . " and error output: " . $this->getErrorOutput());
+                }
+            }
+            return $return_value;
         }
         Log::info("Process not executed in debug mode.");
         return 0;
--- a/app/Console/Commands.php
+++ b/app/Console/Commands.php
@@ -18,7 +18,7 @@ class Commands
         }
 
         $process = Process::fromShellCommandline(config('commands.ping') . " $router->ip -c 1 | grep 'error\|unreachable'");
-        $process->run();
+        $process->run($log = false);
         $result = $process->getOutput(debugOutput: rand(1, 10) > 9 ? "error" : '');
         return $result;
     }
--- a/app/Console/Commands.php
+++ b/app/Console/Commands.php
@@ -13,7 +13,7 @@ class Commands
 {
     public static function pingRouter($router)
     {
-        if (preg_match('/[a-zA-Z]/', $router->ip)) {
+        if (!preg_match('^(?:(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9][0-9]|[0-9])\.){3}(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9][0-9]|[0-9])$', $router->ip)) {
             throw new \InvalidArgumentException("Invalid IP address: " . $router->ip);
         }
 
--- a/app/Http/Controllers/Secretariat/DocumentController.php
+++ b/app/Http/Controllers/Secretariat/DocumentController.php
@@ -158,7 +158,7 @@ class DocumentController extends Controller
         // TODO: figure out result
         Commands::latexToPdf($pathTex, $outputDir);
 
-        if (config('app.debug')) {
+        if (config('app.debug') && !config('commands.run_in_debug')) {
             return $pathTex;
         } else {
             return $pathPdf;
--- a/app/Models/FreePages.php
+++ b/app/Models/FreePages.php
@@ -79,7 +79,7 @@ class FreePages extends Model
      * Wether the free pages are still available.
      * @return bool
      */
-    protected function available()
+    protected function available() : Attribute
     {
         return Attribute::make(
             get: fn () => now()->isBefore($this->deadline)
--- a/app/Models/PrintJob.php
+++ b/app/Models/PrintJob.php
@@ -97,7 +97,7 @@ class PrintJob extends Model
     /**
      * Attribute for the translated cost.
      */
-    public function translatedCost()
+    public function translatedCost() : Attribute
     {
         return Attribute::make(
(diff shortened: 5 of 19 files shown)


horcsinbalint:
I will move most of the changes in migrations to their own PR. Other than them, these changes are quite small so if no issues are raised, I will merge it on Saturday and check if it works.

app/Utils/Process.php
```diff
@@ -7,17 +7,26 @@
...
-            return parent::run($callback, $env);
+            $return_value = parent::run($callback, $env);
+            if($print_to_log) {
+                if($return_value === 0) {
+                    Log::info("Command: " . $this->getCommandLine() . " executed successfully.");
+                } else {
+                    Log::error("Command: " . $this->getCommandLine() . " failed with error code: " . $return_value . "\nWith output: " . $this->getOutput() . " and error output: " . $this->getErrorOutput());
+                }
+            }
+            return $return_value;
```
sszajbely: `$print_to_log` is unnecessarily verbose I think, you could simply call it `$log` or `$verbose`

database/migrations/2019_10_06_212436_create_mac_addresses_table.php
```diff
@@ -19,7 +18,7 @@ public function up()
             $table->unsignedBigInteger('user_id');
             $table->string('mac_address', 17);
             $table->text('comment');
-            $table->enum('state', MacAddress::STATES)->default(MacAddress::REQUESTED);
+            $table->enum('state', ['APPROVED','REJECTED','REQUESTED'])->default('REQUESTED');
```
sszajbely: I would go as far as saying that we shouldn't even use enums in the database at all. I know it's more safe it SQL also checks if they are enums, but the added value is questionable and it needs altering every time a new option is added to the enum.

database/migrations/2019_10_06_224327_create_print_jobs_table.php
```diff
@@ -19,7 +19,7 @@ public function up()
             $table->text('filename');
             $table->text('filepath');
             $table->unsignedBigInteger('user_id')->nullable();
-            $table->set('state', array_map(fn ($state) => $state->value, PrintJobStatus::cases()));
+            $table->set('state', ['QUEUED', 'ERROR', 'CANCELLED', 'SUCCESS']);
```
sszajbely: Same goes for this set, this could just be an array, which can be cast to the set in the model.

database/migrations/2023_01_18_201909_create_voting_tables.php
```diff
@@ -33,13 +29,13 @@ public function up()
         });
         Schema::create('question_options', function (Blueprint $table) {
             $table->id();
-            $table->foreignIdFor(Question::class)->onDelete('cascade');
+            $table->foreignId('question_id')->onDelete('cascade');
```
sszajbely: Do not forget to add the foreign constraint here (see [here](https://laravel.com/docs/11.x/migrations#foreign-key-constraints)), IE. `$table->foreignId('question_id')->constrained()`. Same thing applies to all other cases as well.

Edit: I see now that you have just changed `foreignIdFor` to `foreignId` in every migration file, so where the constraint wasn't enforced before it won't be enforced now. So I don't know if adding it is a good idea then, but it may be worth to consider.

sszajbely (commented):
Otherwise I completely agree with not using models inside migrations, and I feel like Laravel show a wrong path by providing the function `foreignIdFor`.

database/migrations/2019_10_06_212436_create_mac_addresses_table.php
kdmnk: Agreed. Here, we might change it later. 
Also Laravel now can cast DB values to PHP Enums which I believe triggers a similar validation.

app/Models/Semester.php
```diff
@@ -258,29 +258,23 @@ public function questionsNotAnsweredBy(User $user)
...
+            $year = $now->year - 1;
+        } else {
+            $part = 1;
+            // This assumes that the semester ends in the new year.
+            $year = $now->month <= self::END_OF_AUTUMN_SEMESTER ? $now->year - 1 : $now->year;
         }
-
-        return Cache::get('semester.current.' . $today);
+        $current = Semester::getOrCreate($year, $part);
+        return $current;
```
kdmnk: why are you deleting caching?

app/Utils/Process.php
horcsinbalint: Thank you, updated it

database/migrations/2019_10_06_212436_create_mac_addresses_table.php
horcsinbalint: I would create a new, seperate migrations for these in #670 or in a follow-up for it

database/migrations/2023_01_18_201909_create_voting_tables.php
horcsinbalint: Yes, I was also wondering this but decided to leave it without it being `->constrained()`. I would rather add it in a new seperate migration.

database/migrations/2019_10_06_212436_create_mac_addresses_table.php
horcsinbalint: The reason for creating a seperate migration is that I would not like to manually change it on the live system. But before that I will check if the actual live schema is different from what migrations produce for us.

app/Models/Semester.php
horcsinbalint: I forgot to remove it before pushing. It was a pain to debug my changes with rapid fresh generations of the database and having cache (the cache remained intact while I executed `php artisan migrate:fresh --seed`).

I was wondering in the last couple of days how much real world benefits does caching have in similar instances, but sometimes it makes debugging harder (cache integrity has bit me just a couple of days ago 😅 I just started disabling caching temporarily when I suspect that it might interfere with my debugging and re-enable afterwards). But I did not want to bring it to this PR.

database/migrations/2019_10_06_212436_create_mac_addresses_table.php
horcsinbalint: Added to #671

database/migrations/2023_01_18_201909_create_voting_tables.php
horcsinbalint: Added to #671

database/migrations/2019_10_06_224327_create_print_jobs_table.php
horcsinbalint: Added to #671

Issue #527, opened 2024-05-11
Configure migration squashing and make better seeders


kdmnk:
See https://laravel.com/docs/11.x/migrations#squashing-migrations

I think we can't squash the database migrations now because we also insert our constants there and those are ignored when making the squashed file.

Move the default data (roles, etc.?) to a new, not dev-only seeder and enable squashing.

PR #670, opened 2024-12-25
Remove Model references from migrations

horcsinbalint:
## Description

Old migrations have been changed multiple times which is not great practice. Referencing Models in migrations should be avoided as changing the Model might change older migrations. Migrations are usually the "way of propagating changes you make to your Models (adding a field, deleting a model, etc.) into your database schema" (https://docs.djangoproject.com/en/5.1/topics/migrations/). Therefore migrations should never ever reference the Models as their main purpose is to have a migration between two different versions of models. Once a migration is written, they should usually never be edited in the future (unless the language / framework changes the syntax). But functionally a migration should remain the same in its whole life regardless of how the application is developed.

## Related Issue

Continuation of #669

## Changes

Removed references of Models in migrations

## Additional Notes

I will rebase this commit once #669 and #394 is merged, that's why it is a draft


horcsinbalint:
@coderabbitai pause

horcsinbalint:
@coderabbitai full review

database/migrations/2023_05_11_092913_create_study_lines_table.php
```diff
@@ -17,20 +15,22 @@ public function up()
     {
         Schema::create('study_lines', function (Blueprint $table) {
             $table->id();
-            $table->foreignIdFor(EducationalInformation::class);
+            $table->foreignId('educational_information_id');
```
kdmnk, line 18: Make sure that the constraints are still there (https://laravel.com/docs/11.x/migrations#foreign-key-constraints).
But now that I check they don't say that 'foreignIdFor' makes the constraints anyway. Maybe we have set them later?
Check this please throughout the pr.

kdmnk (approved):
Nice, if you are interested, you can also take a look at https://github.com/EotvosCollegium/mars/issues/527
so that our migrations follow the intended behaviour (I think)

database/migrations/2023_05_11_092913_create_study_lines_table.php
horcsinbalint: I compared the production schema and what is produced by this PR and adjusted these migrations to match the production server's schema including table names, field names, field types, constraints, etc. (a1a1e01756ce0def5be2f7c491d39df2578762f8) The schemas are now equivalent.
The schema has some problems (missing some constraints - like the one you mentioned, mismatching datatypes, etc.). I will probably look into it in the future as part of #527 and #671 .
I need the changes from this PR to make the updates for this semester's evaluation form.

PR #674, opened 2024-12-30
Portable devcontainer

horcsinbalint:
This devcontainer setup does not need any kind of dependency to be installed on the development machine except for VS Code, Docker (Compose) unlike the previous dev container setup in this project which required php to be installed on the machine hindering its usability.


BertalanD (approved):
Tried this locally, seems to work fine (macOS host, no php installed locally)

PR #679, opened 2025-01-06
Added Training Codes to be stored, bugfixes

### LINKED ISSUE #529: Store training codes
It has been requested that we should also store training codes (Képzés kód) for students.
It should be collected during the application, and should be available like other information about the educational programme on the user's profile page, and should probably be available through the export feature on the users page.
We should also include an explanation on how users can find this. (Like in Neptun they should go to Studies > Training data [Tanulmányok > Képzés adatok] and they can find it there. And we should also include an example, like "IK-PROGINF-NBHU" for Computer Science BSc.)

horcsinbalint:
## Description
Added Training Codes to study lines and fixed bugs.
## Related Issue
Closes #529.
## Changes
- Added Training Codes to study lines.
- Bugfixes:
  - WorkshopBalances are now only recalculated when the Gazdasági Alelnök initiates this change (it should not be triggered automatically).
  - Decreased the time the current semester is cached to 10 seconds as having cache made debugging extremely difficult. With 10 seconds, the caching has its probable positive effects but it causes interference harder.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,3 +11,10 @@ seed:
 
 serve:
 	php artisan serve --host=0.0.0.0
+
+clear_cache:
+	php artisan cache:clear
+	php artisan view:clear
+	php artisan config:clear
+	php artisan event:clear
+	php artisan route:clear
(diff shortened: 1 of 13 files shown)


horcsinbalint:
The Szakmai Alelnök approved the changes in the perspective of a user

BertalanD (approved):

BertalanD:
Please fix the unit tests:

```
1) Tests\Feature\ApplicationTest::test_submit
Expected response status code [302] but received 500.
Failed asserting that 500 is identical to 302.

The following exception occurred during the last request:

ErrorException: Undefined array key "training_code" in /__w/mars/mars/app/Http/Controllers/Secretariat/UserController.php:[22](https://github.com/EotvosCollegium/mars/actions/runs/12640423722/job/35220811301?pr=679#step:8:23)6
```

PR #681, opened 2025-01-06
Use CLI for semester finalization instead of periodic events

### LINKED ISSUE #680: Do not automatically deactivate collegists when Szemeszter értékelés ends
Szemeszter értékelés should be finalized similarly through CLI like how the admission process is finalized. This request is in line with the requests of the Választmány.

horcsinbalint:
## Description
Szemeszter értékelés should be finalized similarly through CLI like how the admission process is finalized. This PR is in line with the requests of the Választmány.
## Related Issue
Closes #680
## Changes
- Removed automatic deactivation of collegists in the end of Semester evaluation
- Added command for deactivation of collegists for Semester evaluation

### DIFF
--- a/app/Console/Commands/FinalizeSemesterEvaluation.php
+++ b/app/Console/Commands/FinalizeSemesterEvaluation.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+use App\Http\Controllers\Secretariat\SemesterEvaluationController;
+use Illuminate\Support\Facades\App;
+use App\Models\Semester;
+
+class FinalizeSemesterEvaluation extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'app:finalize-semester-evaluation {year} {part}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Finalizes the semester evaluation, deactivates collegists who has not set their new status.';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle()
+    {
+        $year = $this->argument('year');
+        $part = $this->argument('part');
+
+        
+        $semester = Semester::where('year', $year)->where('part', $part)->first();
+
+        App::setLocale('hu');
+        app(SemesterEvaluationController::class)->finalize($semester);
+    }
+}
--- a/app/Http/Controllers/Secretariat/SemesterEvaluationController.php
+++ b/app/Http/Controllers/Secretariat/SemesterEvaluationController.php
@@ -89,7 +89,7 @@ class SemesterEvaluationController extends Controller
     public function handlePeriodicEventReminder(int $daysBeforeEnd): void
     {
         if($daysBeforeEnd < 3) {
-            $userCount = $this->usersHaventFilledOutTheForm()->count();
+            $userCount = SemesterEvaluationController::usersHaventFilledOutTheForm($this->semester())->count();
             Mail::to(config('contacts.mail_membra'))->queue(new EvaluationFormReminder($userCount, $this->getDeadline()));
         }
     }
@@ -97,9 +97,8 @@ class SemesterEvaluationController extends Controller
     /**
      * Send email about results and deactivate collegists who did not fill out the form.
      */
-    public function handlePeriodicEventEnd()
-    {
-        $users = $this->usersHaventFilledOutTheForm();
+    public static function finalize($semester){
+        $users = SemesterEvaluationController::usersHaventFilledOutTheForm($semester);
         $users_names = $users->pluck('name')->toArray();
 
         if (User::secretary()) {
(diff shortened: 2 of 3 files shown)


app/Http/Controllers/Secretariat/SemesterEvaluationController.php
```diff
@@ -94,12 +94,13 @@ public function handlePeriodicEventReminder(int $daysBeforeEnd): void
         }
     }
 
+    public function handlePeriodicEventEnd(){}
```
sszajbely: AFAIK doing nothing is the default behavior of the periodic event trait, so this is not necessary.

app/Console/Commands/FinalizeSemesterEvaluation.php
```diff
@@ -0,0 +1,40 @@
...
+    public function handle()
+    {
+        $year = $this->argument('year');
+        $part = $this->argument('part');
+
+        
+        $semester = Semester::where('year', $year)->where('part', $part)->first();
+
+        App::setLocale('hu');
+        app(SemesterEvaluationController::class)->finalize($semester);
```
sszajbely, line 38: I find this `app(SemesterEvaluationController::class)` call really ugly. I understand the cause, as the periodic events are tied to the controller right now. 
What you are doing is basically finalizing the evaluation given a semester, which in my opinion should not be tied to an instance of the controller. The only thing tying the event to the controller is the name of the controller, and by default the trait's `semester` function (which is the only function from the instance that the `finalize` function uses) returns the semester corresponding to the last periodic event with the controller's name. So I really think this should be a static function of the `SemesterEvaluationController`, which then gets the periodic event by the classname and the semester, finalizing that specific event. I don't know if traits support static functions but this may be solved by rewriting the `periodicEventForSemester` function of the trait to be static.
What do you think @kdmnk? Maybe this is not how you intended periodic events to be used, but you know that I never really liked the idea of tying them to the controllers 😅

app/Http/Controllers/Secretariat/SemesterEvaluationController.php
horcsinbalint: You are right, thanks

app/Console/Commands/FinalizeSemesterEvaluation.php
horcsinbalint: Thanks, I agree with that using an instance of the Controller to change a different semester's evaluation was bad.
I changed `finalize` to be static which solves this issue.
I'm a bit worried on adding heuristics to determine which semester needs to be edited. (We had unintended firings two times last February of the finalization of the end of semester evaluation - even had issues with incorrectly assuming the current semester. That's why I wanted it to be triggered by human action as it can be quite destructive. With this, the deadline can also be extended after it has expired and the data can be manually corrected before any destructive action is called. These were features that the Student Council also wanted.) I think getting the semester through CLI makes very little room of error.

kdmnk: Have you had unintended firings since the periodic events were introduced? "incorrectly assuming the current semester" seems strange anyways as periodic events are not based on the current semester.

But I need to get my head around it and merge the other periodic event relevant PRs that I'm delayed with. 

My only problem here is that the periodic event stores and encapsulates the corresponding semester. Providing a semester by an argument will mess that up. But again, I will get back to this at the weekend I think, if thats not too late.

horcsinbalint: There was a functional requirement from @kamillacziraki to be able to reopen the form after its deadline. So setting manual firing is a functional requirement. It is not possible to do in response to an HTTP request as quite a few emails are sent out, so processing it can take a significant amount of time. The processing time for HTTP requests might be limited in accordance with industry standards. So manual firing can only be implemented through using CLI.

We (sysadmins) have received a request to reopen the form at the end of last semester. At the end the request was rejected as doing a partial rollback of the database would have been more effort than the benefits it would have offered. Doing a single partial rollback once takes more effort than it was developing the CLI tool proposed in this PR and using the CLI tool 20 times.

Periodic events were firing exactly at times they were set to.

> "incorrectly assuming the current semester" seems strange anyways as periodic events are not based on the current semester.

There is no active periodic event to get the data from if the semester finalization is initiated manually after the deadline. Using the most recent, closed periodic event and not ask it via an argument is overengineering IMO and can cause to incorrect semester to be selected (like if another semester's form was also opened recently by accident).
If someone is knowledgeable enough to execute a CLI command, they should be able to specify which semester's form to be closed.

Similar solution is used to finalize the admission process (a command needs to be executed through CLI), and works perfectly.

sszajbely: > It is not possible to do in response to an HTTP request as quite a few emails are sent out, so processing it can take a significant amount of time. The processing time for HTTP requests might be limited in accordance with industry standards. So manual firing can only be implemented through using CLI.

This may not be relevant at all, but this is where you would normally use Jobs, to do work in a background thread. I'm not sure if you are unaware or something else, but I thought it was worth mentioning.

horcsinbalint: Thanks for pointing it out, created an issue for it (#685).

BertalanD (approved):

Issue #661, opened 2024-11-22
Create end-of-semester form for seniors


horcsinbalint:
Currently, seniors need to submit their report in PDF format by e-mail. It would probably be better if they could submit these PDFs via mars.

The challenges involved / please note:
- this depends on #660
- it should be easy to download all reports at once by the people involved (e. g. downloading multiple files requires permissions from the browser, adding a dependency for zipping files might cause security risks)
- the deadline for these might be different from the forms provided for non-seniors
- the director might disagree with this request so it probably needs not to be implemented (this issue is only created to track this feature request that I received from the Student Council)
- seniors might not be required to fill out a form, they are only required to submit their report (i. e. the system might not have the authority to force them to answer if they want to continue with their senior status before they can submit their report)

PR #682, opened 2025-01-07
Added senior role and fixed problems with end-of-semester form

### LINKED ISSUE #660: Problem with seniors and end-of-semester form
Seniors should not fill out the end-of-semester form but in the current implementation they have the ability to fill it out and they also receive misleading e-mails about these forms. To resolve this issue, the following changes should be implemented:

- create a senior role
- collegists with senior role should not be required to fill out the form (e. g. should be activated for the next semester even if they do not fill out the form)
- collegists with senior role should not be able to fill out the form
- fix all misleading texts in blades, emails, etc. where a senior is asked to fill out the form
- search for other inconsistencies regarding seniors

As this issue is quite profound and needs to be fixed before this semester's form is sent out, I will take on this issue on the 21st of December, 2024.

horcsinbalint:
## Description
There have been multiple issues because mars did not have a registry of who is a senior. Especially in the case of the end-of-semester form as they are governed by a completely seperate ruleset as other collegists.
## Related Issue
Closes #660
## Changes
- Created senior role
- Seniors cannot and do not have to fill out the end-of-semester form
- Fixed email texts that seniors do NOT have to fill out the end-of-semester form and their report is not part of mars currently

### DIFF
--- a/app/Http/Controllers/Secretariat/SemesterEvaluationController.php
+++ b/app/Http/Controllers/Secretariat/SemesterEvaluationController.php
@@ -137,7 +137,6 @@ class SemesterEvaluationController extends Controller
         $this->authorize('fillOrManage', SemesterEvaluation::class);
 
         return view('secretariat.evaluation-form.app', [
-            'phd' => user()->educationalInformation->isSenior(),
             'user' => user(),
             'faculties' => Faculty::all(),
             'workshops' => Workshop::all(),
--- a/app/Http/Controllers/Secretariat/SemesterEvaluationController.php
+++ b/app/Http/Controllers/Secretariat/SemesterEvaluationController.php
@@ -245,9 +245,15 @@ class SemesterEvaluationController extends Controller
      */
     public function usersHaventFilledOutTheForm()
     {
-        return User::withRole(Role::COLLEGIST)->verified()->whereDoesntHave('semesterStatuses', function ($query) {
-            $query->where('semester_id', $this->semester()?->succ()?->id);
-        })->get();
+        return User::withRole(Role::COLLEGIST)
+            ->whereDoesntHave('roles', function ($query) {
+                $query->where('name', Role::SENIOR);
+            })
+            ->verified()
+            ->whereDoesntHave('semesterStatuses', function ($query) {
+                $query->where('semester_id', $this->semester()?->succ()?->id);
+            })
+            ->get();
     }
 
     /**
--- a/app/Models/EducationalInformation.php
+++ b/app/Models/EducationalInformation.php
@@ -8,6 +8,7 @@ use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\HasMany;
 
 use App\Models\Semester;
+use App\Models\Role;
 
 /**
  * Either B2 or C1.
@@ -113,11 +114,11 @@ class EducationalInformation extends Model
     }
 
     /**
-     * Whether the user is a senior (i.e. currently has a PhD study line).
+     * Whether the user is a senior
      */
     public function isSenior(): bool
     {
-        return $this->studyLines()->currentlyEnrolled()->where('type', 'phd')->exists();
+        return $this->user->isSenior();
     }
 
     /**
--- a/app/Models/Role.php
+++ b/app/Models/Role.php
@@ -51,6 +51,7 @@ class Role extends Model
     public const ETHICS_COMMISSIONER = 'ethics-commissioner';
     public const ALUMNI = 'alumni';
     public const RECEPTIONIST = 'receptionist';
+    public const SENIOR = 'senior';
 
     //Students' Committe role's objects
     public const PRESIDENT = 'president';
@@ -125,7 +126,8 @@ class Role extends Model
         self::BOARD_OF_TRUSTEES_MEMBER,
         self::ETHICS_COMMISSIONER,
         self::ALUMNI,
-        self::RECEPTIONIST
+        self::RECEPTIONIST,
+        self::SENIOR
     ];
 
     protected $fillable = [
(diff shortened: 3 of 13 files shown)


horcsinbalint:
@kamillacziraki what do you think about the text of the e-mails?

kamillacziraki:
> @kamillacziraki what do you think about the text of the e-mails?

I like them, I think it will be okay this way. If all the seniors send their reports on time, we will be able to manage them significantly better than before.

kdmnk:
We do know who is a senior. It is calculated based on the study lines in the EducationalInformation. 
Seniors still had to fill a shortened end of semester report, just to keep track of their status (active, erasmus, resident etc) but their form did not include grades and such.

What was the problem with this?

kamillacziraki:
> We do know who is a senior. It is calculated based on the study lines in the EducationalInformation. Seniors still had to fill a shortened end of semester report, just to keep track of their status (active, erasmus, resident etc) but their form did not include grades and such.
> 
> What was the problem with this?

The main problem was that there are usually a lot of seniors who don't fill it out at all, resulting in them being declared alumni in Urán, but, as they send in their senior reports as usual, they are included in the Secretariat's list of seniors. According to the CTSZK, they only have to send in their senior report, so it wasn't clear why they have to do the form as well, and recently, this caused a lot of problems with the seniors' records. This was why I asked Bálint if we can disable the report for them, even if that means we have to update their status manually for now (on the long run, I think it would be great to have the senior reports integrated, so that they have to upload it directly to Urán, that would be the ideal solution, but until then, this makes it easier to keep track of them).

sszajbely:
> > We do know who is a senior. It is calculated based on the study lines in the EducationalInformation. Seniors still had to fill a shortened end of semester report, just to keep track of their status (active, erasmus, resident etc) but their form did not include grades and such.
> > What was the problem with this?
> 
> The main problem was that there are usually a lot of seniors who don't fill it out at all, resulting in them being declared alumni in Urán, but, as they send in their senior reports as usual, they are included in the Secretariat's list of seniors. According to the CTSZK, they only have to send in their senior report, so it wasn't clear why they have to do the form as well, and recently, this caused a lot of problems with the seniors' records. This was why I asked Bálint if we can disable the report for them, even if that means we have to update their status manually for now (on the long run, I think it would be great to have the senior reports integrated, so that they have to upload it directly to Urán, that would be the ideal solution, but until then, this makes it easier to keep track of them).

In this case maybe the solution is to modify the CTSzK instead of going backwards and hot fixing the issue. I'm quite unsure how you would want the senior reports to be implemented into Urán when right now this is exactly the opposite you are proposing. This is exactly the same kind of problem that we hoped to fix with the ruleset changes proposed two years ago, explicitly mentioning Urán in the rulesets, thus making the use of it officially required.

I do understand that adding a role to Urán and "solving" the problem here is much easier than modifying any rulesets, but I really do find it bad practice that everyone tries to go the easy way instead of the right way. This is quite clear by how in the two years since the last ruleset modifications (to SzMSz) were proposed, there was not one member of the student council who tried to make the Senate sign off on it.

kamillacziraki:
> In this case maybe the solution is to modify the CTSzK instead of going backwards and hot fixing the issue. I'm quite unsure how you would want the senior reports to be implemented into Urán when right now this is exactly the opposite you are proposing. This is exactly the same kind of problem that we hoped to fix with the ruleset changes proposed two years ago, explicitly mentioning Urán in the rulesets, thus making the use of it officially required.
> 
> I do understand that adding a role to Urán and "solving" the problem here is much easier than modifying any rulesets, but I really do find it bad practice that everyone tries to go the easy way instead of the right way. This is quite clear by how in the two years since the last ruleset modifications (to SzMSz) were proposed, there was not one member of the student council who tried to make the Senate sign off on it.

I agree that right now this is a hot fix rather than a permanent solution, but I think that keeping the report the same while we work on the long-term solution is not a good idea since it causes a lot of confusion and double work. Later on, of course we could (and should) specify Urán as officially required in this case as well, but since this can't be achieved for this semester, we agreed with Bálint that it would be best that while we work on that, we make this change.

horcsinbalint:
> We do know who is a senior. It is calculated based on the study lines in the EducationalInformation.

Thanks for pointing out, I removed it so that it would not be redundant.

It would be nice to be able to filter for seniors in the users page (@kamillacziraki and others also requested it). The current system practically allows users to serve themselves with something similar to a senior role. This seems to be a bit excessive permission. It should be granted by someone who is in charge of maintaining the roles.
They cannot just simply set themselves to be a resident either.
Another theoretical problem that could happen if someone is getting their PhD non-related to the College and are admitted to the College as doing a BA/BSc. The active code would consider them as seniors. Yeah, an edge-case, but I don't think there is any way of storing their data in a appropriately (that works correctly with any kind of filtering) without developing anything in mars.
Adding senior role to a few (4-8) collegists yearly is not a big effort and I think this is a more correct way of implementing this.

> explicitly mentioning Urán in the rulesets, thus making the use of it officially required

Nothing legally prevents us from collecting the senior reports through Urán. But the current questions simply do not meet the standards of their reports that is requested from seniors.
Their report also might have a different deadline than other collegists' the end-of-semester form.
There is an issue that has specified the requirements to move the report of seniors into mars (#661). It is simply not going to be implemented prior to this semester's evaluation period as nobody took on #661 so even if it is implemented in a few days, I cannot test it thoroughly.

> hot fixing the issue

I debate that this is a hotfix of the issue. The issue that had a list of functional changes (#660) was opened on Nov 22, 2024 giving plenty of time to discuss the functional requirements.

While I can see why there is discussion on adding a senior role as it is a technical question - and I also thank for the comment -, I do not fully understand why political issues are discussed in this much depth under the pull request.

BertalanD:
> Nothing legally prevents us from collecting the senior reports through Urán. But the current questions simply do not meet the standards of their reports that is requested from seniors.

And that's not necessarily a problem; it's explicitly desired that seniors provide a long-form textual report that is individually reviewed by the director. We could technically create a file upload for that, but I think the status quo is perfectly manageable considering the small number of seniors (as long as they follow the instructions and don't forget to cc `elnok@`)

kdmnk (approved):

Issue #689, opened 2025-01-09
Printing from `DocumentController` is broken


BertalanD:
@sszajbely 

The logic in `DocumentController::printDocument` seems to have been missed in #394, and the function references the now-nonexistent `App\Utils\Printer` class:

https://github.com/EotvosCollegium/mars/blob/4bfe84d7b8ca37e0d5cb4f21961fdc955f084a44/app/Http/Controllers/Secretariat/DocumentController.php#L135-L143

We should factor out the code for checking and deducting print balances and initiating the print job from `PrintJobController::store` into a function that can be called from backend logic.

sszajbely:
Sorry about this oversight. I think the `PrinterHelper` class would be the most suitable with a function like
```php
function tryCreatePrintJob(string $documentPath, bool $doubleSided, <etc.>) PrintJobCreationProcess
``` 
where `PrintJobCreationProcess` is an enum with values like `PrintJobCreationProcess::noBalance`, `PrintJobCreationProcess::errorPrinting` or `PrintJobCreationProcess::success`. Then both the `store` method of the `PrintJobController` and the `DocumentController` can use this function and based on the returned enum value send a message to the user.

PR #690, opened 2025-01-09
Resolve most CI problems

BertalanD:
The first commit adds PHP CS Fixer as a dev dependency and configures the dev container to use it for formatting PHP code. If we make use of this tool, there will be less noise created by DeepSource Autofix.

The second fix resolves most phpstan analysis issues that trigger on master. The single remaining problem is tracked in #689

### DIFF
--- a/app/Models/PrintAccount.php
+++ b/app/Models/PrintAccount.php
@@ -69,7 +69,7 @@ class PrintAccount extends Model
 
     /**
      * The free pages which are currently available. Sorts the free pages by their deadline.
-     * @return Collection
+     * @return Collection|FreePages[]
      */
     public function availableFreePages()
     {
@@ -131,7 +131,7 @@ class PrintAccount extends Model
             $availableFreePages = $this->availableFreePages()->where('amount', '>', 0);
 
             // Subtract the pages from the free pages pool, as many free pages as necessary
-            /** @var FreePages */
+            /** @var FreePages $freePages */
             foreach ($availableFreePages as $freePages) {
                 $subtractablePages = $freePages->calculateSubtractablePages($freePagesToSubtract);
                 $freePages->subtractPages($subtractablePages);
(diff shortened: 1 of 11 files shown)


app/Models/Printer.php
```diff
@@ -108,7 +108,8 @@ public function print(bool $twoSided, int $copies, string $path)
             throw new PrinterException($e->getMessage(), $e->getCode(), $e->getPrevious());
         }
 
-        return $jobId;
+        // CUPS job IDs are integers (see https://www.cups.org/doc/cupspm.html#cupsCreateDestJob)
+        return (int)$jobId;
```
horcsinbalint: I think lp contains non-numeral characters as a job id. I think I had to change exactly this line.

horcsinbalint (changes requested):
lp gives out a string and not an int if I know it correctly

app/Models/Printer.php
horcsinbalint: I had to remove the int cast in a05f8e434402b5ae23ea06719cb0f72d209a49e9, as the jobId received from lp is a string that contains alpha characters

app/Models/PrintAccount.php
```diff
@@ -69,7 +69,7 @@ public function freePages()
 
     /**
      * The free pages which are currently available. Sorts the free pages by their deadline.
-     * @return Collection
+     * @return Collection|FreePages[]
```
horcsinbalint, line 72: Cannot it be simply `@return FreePages[]`?

horcsinbalint (approved):

app/Models/PrintAccount.php
BertalanD: This is _never_ actually an array of `FreePages` objects; however PHPStan is not able to figure out the element type of the `Collection` otherwise. This workaround is suggested in [this StackOverflow answer](https://stackoverflow.com/a/70699216) and is already present elsewhere in the code.

PR #691, opened 2025-01-10
Remove unused Vue dependency

BertalanD:
`vue-template-compiler` has been flagged by Dependabot as vulnerable to XSS. This made me realize that this dependency is not used for anything, so let's yeet it.


horcsinbalint (approved):
LGTM

PR #696, opened 2025-01-31
Use the PeriodicEvent's semester() for semester evaluation on most occasions

horcsinbalint:
This is quite a bit of hack and should be cleaned up


horcsinbalint:
The failing unit tests are just caused by the current date, the older commit also failed now after rerun

BertalanD (approved):
this definitely needs to be fixed properly (maybe bundle data that is only relevant for active semesters into a single top-level object that might be null). But for now, this works and fixes UI bugs (previously saved values are not reflected as data is fetched from the wrong semester)

PR #711, opened 2025-02-13
Adding an STV question to the seeder + count of abstentions to details

viktorcsimma:
The only problem is that votes are not linked to any user; so the number of ballots is not correct.

See the blockquote under the _Details_ part for those enhancements.

grafik


horcsinbalint (changes requested):
This PR has bugs, won't merge today

PR #712, opened 2025-02-15
Use free printing credits instead of free printing pages

horcsinbalint:
Moving from free printing pages to free printing credits. Printing credit usage is calculated the same way as the normal printing account.


horcsinbalint:
@coderabbitai review

resources/views/dormitory/print/manage/account_history.blade.php
```diff
@@ -26,8 +26,8 @@
                 minWidth:100
             },
             {
-                title: "Ingyenes oldal változás",
-                field: "free_page_change",
+                title: "Ingyenes nyomtatási kredit változás",
+                field: "free_printing_credits_change",
```
BertalanD, line 30: I assume you tested this thoroughly and everything works, however, inconsistently calling this `free_printing_credit_change`/`free_printing_credits_change` (notice the plural!) might cause problems later on.

resources/views/dormitory/print/free.blade.php
```diff
@@ -25,10 +25,10 @@
             columnMinWidth: 150,
             headerSort: false,
             ajaxURL: "{{ $route }}", //set url for ajax request
-            placeholder: "@lang('print.no_free_pages')",
+            placeholder: "@lang('print.no_free_printing_credits')",
             columns: [
                 @if($admin)
-                @can('viewAny', App\Models\FreePages::class)
+                @can('viewAny', App\Models\FreePrintingCredits::class)
```
BertalanD, line 30: the closing `@endif` and `@endcan` tags are in the wrong order in lines 45 and 46. How does this work?

database/migrations/2025_02_15_113728_rename_free_pages_to_credits.php
```diff
@@ -0,0 +1,35 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::rename("printing_free_pages", "printing_free_printing_credits");
```
BertalanD, line 13: Do I understand correctly that this will change my existing 74 free pages into 74 HUF (i.e. 7 pages). Wouldn't it make more sense to drop the entire table and re-assign the appropriate amount of credits for users?

resources/views/dormitory/print/manage/account_history.blade.php
horcsinbalint: It did make the user page not load. Fixed it, thanks for pointing out

resources/views/dormitory/print/free.blade.php
horcsinbalint: According to my testing, `@endcan` is the same as `@endif`, though I'm not really sure. But checking if the user is admin seems to be redundant, removed the check. Thank you for pointing out.

database/migrations/2025_02_15_113728_rename_free_pages_to_credits.php
horcsinbalint: Yep, it does exactly that. I did not want to delete old records. Essentially we are just renaming free pages to free credits, and then increase the price of printing if someone pays with credits. Price increases may happen in the future. The 900% inflation is a bit extreme though. I'm planning on removing quotas by hand that are erroneous due to this change and give out new quotas.

app/Models/PrintJob.php
```diff
@@ -134,18 +134,6 @@ public function cancel()
...
-
-                if ($this->used_free_pages) {
-                    $pages = $printAccount->availableFreePages()->first();
-                    $pages->update([
-                        'last_modified_by' => user()->id,
-                        'amount' => $pages->amount + $this->cost,
-                    ]);
-                } else {
-                    $printAccount->balance += $this->cost;
-                }
```
BertalanD, line 148: your call, but probably deserves some explanation

database/migrations/2025_02_15_113728_rename_free_pages_to_credits.php
horcsinbalint: So I think it would make sense to truncate the table, but I don't think it is necessary and intended to fix the issues with it manually. I checked it before opening the PR, only a few users (~10-20) has free pages that is not expired, so we can adjust it manually.

BertalanD (approved):

database/migrations/2025_02_15_113728_rename_free_pages_to_credits.php
sszajbely: I don't think this approach is particularly clean. I see your point in keeping the records, and I agree with it, but I think the clean approach would be to:
1. Create the new table with the appropriate fields.
2. Copy all information from the old table to the new table.
3. Drop the old table.

And while you do this, you can also simply multiply the number of pages by the cost per page. Like this, you keep the records, which are then also more readable for users, as there won't be strange records where people have seemingly been printing for 1 Ft per page. And as no inflation happened in the past years (at least to my knowledge, but this maybe not be true anymore), you wouldn't even create false information by simply multiplying the pages by the cost per page.

app/Models/PrintJob.php
horcsinbalint: I think if you start a printing job, and you cancel it just at the last second of printing, you would get your money back and also have the printed document. (Note that a `PrintJobStatus`es can only be `QUEUED`, `CANCELLED`, `SUCCESS`. There is no such state that would indicate that it is being printed.) Users should reach out to system administrators if they need reimbursement. (We might add a button in the future where they can flag a PrintJob as erroneously billed.)
I tried to demonstrate it, but cancellation does not even work at all: #713

database/migrations/2025_02_15_113728_rename_free_pages_to_credits.php
horcsinbalint: You convinced me, I added the multiplication to the migration (and division for the case of reverting the migration). I think renaming with `Schema::rename` is cleaner than manually copying the table because it does not break foreign keys that reference a table (there are probably no references here).

PR #724, opened 2025-04-21
Allow users to opt out of Mr and Miss participation

BertalanD:
We show a simple button a week before the voting period's start where
users can declare that they do not want to participate. In that case,
they won't show up in the drop-down lists for other users, and they
themselves can't vote either.

### DIFF
--- a/app/Http/Controllers/StudentsCouncil/MrAndMissController.php
+++ b/app/Http/Controllers/StudentsCouncil/MrAndMissController.php
@@ -7,9 +7,11 @@ use App\Models\MrAndMissCategory;
 use App\Models\MrAndMissVote;
 use App\Models\Semester;
 use App\Models\User;
+use App\Models\MrAndMissOptOut;
 use App\Utils\HasPeriodicEvent;
 use Carbon\Carbon;
 use Illuminate\Auth\Access\AuthorizationException;
+use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\DB;
@@ -19,12 +21,12 @@ class MrAndMissController extends Controller
     use HasPeriodicEvent;
 
     /**
-     * Show the page for voting.
+     * Show the page for voting and opting out.
      * Accessible by collegists.
      */
     public function index(Request $request)
     {
-        $this->authorize('voteOrManage', MrAndMissVote::class);
+        $this->authorize('accessOrManage', MrAndMissVote::class);
 
         $categories = MrAndMissCategory::select(['mr_and_miss_categories.id', 'title', 'mr', 'custom', 'votee_id', 'votee_name as custom_name'])
             ->where('hidden', false)
@@ -46,9 +48,11 @@ class MrAndMissController extends Controller
             'student-council.mr-and-miss.index',
             [
                 'categories' => $categories,
-                'users' => User::collegists(),
+                'users' => $this->participatingUsers(),
                 'miss_first' => rand(0, 1) == 0,
                 'deadline' => $this->getDeadline(),
+                'start_date' => $this->getStartDate(),
+                'opted_out' => $this->optedOut(user()),
             ]
         );
     }
@@ -149,7 +153,7 @@ class MrAndMissController extends Controller
     {
         $this->authorize('manage', MrAndMissVote::class);
 
-        if(!$this->semester()) {
+        if (!$this->semester()) {
             throw new \Exception('Nincs megjeleníthető eredmény.');
         }
         $results = MrAndMissVote::select(DB::raw('count(*) as count, users.name, votee_name, title, mr, custom'))
@@ -230,5 +234,60 @@ class MrAndMissController extends Controller
         return redirect()->back()
             ->with('activate_custom', 'true') //go to the custom category page
             ->with('message', __('general.successful_modification'));
+    }
+
+    /*
+     * Opt out of participating.
+     * Accessible by collegists in the week leading up to the voting period.
+     */
+    public function optOut(Request $request)
+    {
+        $this->authorize('optOut', MrAndMissVote::class);
+
+        $validatedData = $request->validate([
+            'opt_out' => 'required|boolean',
+        ]);
+
+        if ($validatedData['opt_out']) {
+            MrAndMissOptOut::updateOrCreate([
+                'user_id' => user()->id,
+                'semester_id' => $this->semester()->id,
+            ]);
+        } else {
+            MrAndMissOptOut::where([
+                'user_id' => user()->id,
+                'semester_id' => $this->semester()->id,
+            ])->delete();
+        }
+
+        return back()->with('message', __('general.successful_modification'));
+    }
+
+    public function isOptOutPeriod(): bool
+    {
+        $startDate = $this->periodicEvent()?->startDate();
+
+        return $startDate && $startDate->between(Carbon::now(), Carbon::now()->addDays(7));
+    }
+
+    public function optedOut(User $user): bool
+    {
+        return MrAndMissOptOut::where([
+            'user_id' => $user->id,
+            'semester_id' => $this->semester()?->id,
+        ])->exists();
+    }
+
+    private function participatingUsers(): Collection
+    {
+        if (!$this->semester()) {
+            return new Collection();
+        }
+
+        return User::collegist()->active()->whereNotIn('id', function ($query) {
+            $query->select('user_id')
+                ->from('mr_and_miss_opt_outs')
+                ->where('semester_id', $this->semester()->id);
+        })->get();
     }
 }
--- a/app/Models/MrAndMissOptOut.php
+++ b/app/Models/MrAndMissOptOut.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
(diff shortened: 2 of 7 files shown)


app/Http/Controllers/StudentsCouncil/MrAndMissController.php
```diff
@@ -231,4 +235,55 @@ public function customVote(Request $request)
...
+        return $startDate && $startDate->between(Carbon::now(), Carbon::now()->addDays(7));
+    }
+
+    public function optedOut(User $user): bool
+    {
+        return MrAndMissOptOut::where([
+            'user_id' => $user->id,
+            'semester_id' => $this->semester()->id,
+        ])->exists();
+    }
```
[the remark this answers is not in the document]
BertalanD: If `semester` is null, then the permission validation fails even before we'd get here.

app/Http/Controllers/StudentsCouncil/MrAndMissController.php
```diff
@@ -231,4 +235,55 @@ public function customVote(Request $request)
...
+        $this->authorize('optOut', MrAndMissVote::class);
+
+        $validatedData = $request->validate([
+            'opt_out' => 'required|boolean',
+        ]);
+
+        if ($validatedData['opt_out']) {
+            MrAndMissOptOut::updateOrCreate([
+                'user_id' => user()->id,
+                'semester_id' => $this->semester()->id,
```
horcsinbalint, line 254: `$this->semester()` can be null if this is the first Mr&Miss vote.
Probably should be:
`'semester_id' => $this->semester()?->id,`

app/Http/Controllers/StudentsCouncil/MrAndMissController.php
```diff
@@ -231,4 +235,55 @@ public function customVote(Request $request)
...
+            ]);
+        } else {
+            MrAndMissOptOut::where([
+                'user_id' => user()->id,
+                'semester_id' => $this->semester()->id,
```
horcsinbalint, line 259: Same problem with semester being able to be null

app/Http/Controllers/StudentsCouncil/MrAndMissController.php
```diff
@@ -231,4 +235,55 @@ public function customVote(Request $request)
...
+        $startDate = $this->periodicEvent()?->startDate();
+
+        return $startDate && $startDate->between(Carbon::now(), Carbon::now()->addDays(7));
+    }
+
+    public function optedOut(User $user): bool
+    {
+        return MrAndMissOptOut::where([
+            'user_id' => $user->id,
+            'semester_id' => $this->semester()->id,
```
horcsinbalint: Same problem with semester being able to be null

app/Http/Controllers/StudentsCouncil/MrAndMissController.php
```diff
@@ -231,4 +235,55 @@ public function customVote(Request $request)
...
+        ])->exists();
+    }
+
+    private function participatingUsers(): Collection
+    {
+        return User::collegist()->active()->whereNotIn('id', function ($query) {
+            $query->select('user_id')
+                ->from('mr_and_miss_opt_outs')
+                ->where('semester_id', $this->semester()->id);
```
horcsinbalint, line 290: Same problem with semester being able to be null

app/Policies/MrAndMissVotePolicy.php
```diff
@@ -15,6 +16,38 @@ class MrAndMissVotePolicy
...
+    public function optOut(User $user)
+    {
+        if (!($user->isCollegist(alumni: false))) {
+            return Response::deny('Csak collegisták vehetnek részt.');
+        }
+        if (app(MrAndMissController::class)->isActive()) {
+            return Response::deny('A szavazás már zajlik.');
+        }
+        if (!app(MrAndMissController::class)->isOptOutPeriod()) {
+            return Response::deny('A szavazás jelenleg nem elérhető.');
```
horcsinbalint: Is this error message correct?

app/Policies/MrAndMissVotePolicy.php
```diff
@@ -15,6 +16,38 @@ class MrAndMissVotePolicy
...
+    *
+    * @param User $user
+    * @return Response
+    */
+    public function optOut(User $user)
+    {
+        if (!($user->isCollegist(alumni: false))) {
+            return Response::deny('Csak collegisták vehetnek részt.');
+        }
+        if (app(MrAndMissController::class)->isActive()) {
```
horcsinbalint: I think this check is redundant. I also think it is a bit misleading that there is an `isOptOutPeriod` function in `MrAndMissController` and we add further restrictions to it.

app/Policies/MrAndMissVotePolicy.php
```diff
@@ -15,6 +16,38 @@ class MrAndMissVotePolicy
...
+        return $this->vote($user);
+    }
+
+    /**
+    * Determine whether the user can opt out.
+    *
+    * @param User $user
+    * @return Response
+    */
+    public function optOut(User $user)
```
horcsinbalint: nit: could you please also set the return value type like:
`public function optOut(User $user): Response`

app/Models/MrAndMissOptOut.php
```diff
@@ -0,0 +1,25 @@
...
+ * @method static \Illuminate\Database\Eloquent\Builder|MrAndMissOptOut newModelQuery()
+ * @method static \Illuminate\Database\Eloquent\Builder|MrAndMissOptOut newQuery()
+ * @method static \Illuminate\Database\Eloquent\Builder|MrAndMissOptOut query()
+ * @method static \Illuminate\Database\Eloquent\Builder|MrAndMissOptOut whereUserId($value)
+ * @method static \Illuminate\Database\Eloquent\Builder|MrAndMissOptOut whereSemesterId($value)
+ * @mixin \Eloquent
+ */
+class MrAndMissOptOut extends Model
+{
+    public $timestamps = false;
```
horcsinbalint: I would probably include timestamps. (Not much data, and sometimes they can be useful.)

database/migrations/2025_04_21_112234_add_mr_and_miss_opt_out.php
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('mr_and_miss_opt_outs', function (Blueprint $table) {
+            $table->id();
+            $table->unsignedBigInteger('user_id');
+            $table->unsignedSmallInteger('semester_id');
+
```
horcsinbalint, line 17: Timestamps should be also added here if we decide to collect them

horcsinbalint (changes requested):
Thank you very much, I like the change that users would be able to opt out from Mr&Miss. The changes are well made and well commented.
I had some issues with setting it up for the first time though, could you please check the comments that I added on Github? I would also probably include timestamps to be saved for opt outs.
Other than these minor remarks, LGTM.

app/Http/Controllers/StudentsCouncil/MrAndMissController.php
BertalanD: > $this->semester() can be null if this is the first Mr&Miss vote.

Did you actually run into a null deref here? Worked fine for me, and I was under the assumption that `$this->semester()` was set when the periodic event for the voting period is created. (And if there is no periodic event, the `can('optOut')` check aborts the request even before we get here.)

app/Policies/MrAndMissVotePolicy.php
BertalanD: Was trying to avoid translating "opting out" into Hungarian. The text in the UI is already awkward, and I have even less opportunity to write a longer-form explanation.

app/Models/MrAndMissOptOut.php
BertalanD: I have no strong opinion on this, let's just include it then.

app/Http/Controllers/StudentsCouncil/MrAndMissController.php
horcsinbalint: Well, in this case, it shouldn't be a problem as you pointed out. Some of the other similar comments are problematic though

app/Policies/MrAndMissVotePolicy.php
horcsinbalint: Couldn't we just simply give a bool value as a result? This function is only called in case of a POST request (if I know it correctly), so normal users would not be able to see this error message. (Unless they have opened the form before the opt out period ended, but it is quite unlikely.)

horcsinbalint: But we can leave it as it is, it's not a big deal

app/Http/Controllers/StudentsCouncil/MrAndMissController.php
BertalanD: This did actually cause problems; fixed.

app/Policies/MrAndMissVotePolicy.php
BertalanD: How about this?

app/Policies/MrAndMissVotePolicy.php
BertalanD: Changed the entire class to use type annotations instead of doc comments.

horcsinbalint (approved):
LGTM

PR #725, opened 2025-04-24
Add missing foreign key constraints

BertalanD:


horcsinbalint (approved):
Thanks, LGTM, this has bothered me for quite some time.

One small thing that could probably be also done: `semester_id` could be of `unsignedBigInteger` everywhere as it is more commonly used with ids. But if you don't want to increase the scope of the PR, merge it as is. (We should try it on staging before putting it on prod as our database might be inconsistent as there were no foreign keys on these before.)

Issue #486, opened 2024-03-30
Router pinging is using an unstable interface


horcsinbalint:
https://github.com/EotvosCollegium/mars/blob/c90d73a75bb5f9429b7cfd0d6b073152631861d3/app/Console/Commands.php#L75C1-L75C106 uses the `ping` command's output to determine whether a router is available. If a router is available, then it works fine but if a router is not available, sometimes it might misinterpret the results as the `ping` command might give out failed results without those words present as per https://github.com/iputils/iputils/blob/294a65f00047f31685b7dd07048b48c4be82e097/ping/ping.c#L1236
It is not just some theoretical issue, as it has sent a couple of emails because of one router.

horcsinbalint:
We should probably use `nmap` for it as it has multiple output formats that can be easily processed. Probably all routers can be pinged with a single command. For example `nmap IPADDRESSHERE -sP -oG -` can be easily processed (like it contains the IP address and the `Status: Up` string if the pinging is successful and it can process multiple IP-s in a single command).

horcsinbalint:
Fixed by d7c16cd

PR #726, opened 2025-05-18
Remove string processing done via shell and #486

horcsinbalint:
Using shell scripts to transform text contributes quite a bit to the smell of the codebase. We also had issues because of this (`grep` arbitrarily decided sometimes for outputs of `pdfinfo` that they were binary data and not text). While removing calls to `Process::fromShellCommandline` (which is quite unsafe), I also moved from `ping` to `nmap` (fixing https://github.com/EotvosCollegium/mars/issues/486).


app/Utils/PrinterHelper.php
```diff
@@ -19,7 +19,7 @@ class PrinterHelper
      */
     public static function getDocumentPageNumber(string $path): int
     {
-        $process = Process::fromShellCommandline(config('commands.pdfinfo') . " $path | grep '^Pages' | awk '{print $2}'");
+        $process = Process::fromShellCommandline(config('commands.pdfinfo') . " $path | grep --text '^Pages' | awk '{print $2}'");
```
BertalanD: What. Why can't we just simply parse the command's output from within PHP. Opening a shell for this is very fishy.

horcsinbalint: Thanks for the suggestion, I think I have remedied the issue (and some similar as well)

BertalanD (commented):
Thank you, LGTM! (Assumimg this has been tested)

BertalanD (approved):

PR #728, opened 2025-05-18
Readd missing foreign key constraints

horcsinbalint:
#725 after the data has been fixed


BertalanD:
What's the status on this?

horcsinbalint:
I added some cascading and purged the junk from the prod DB. Should be probably good now

PR #731, opened 2025-05-20
Added selectable printers

### LINKED ISSUE #375: Speed up printer page
The now used lpstat can be a bit slow, and we call it many times. Try to optimise it.

Copy of https://github.com/luksan47/mars/issues/120

### LINKED ISSUE #376: Make "flip on short edge" an option when printing
Copy of https://github.com/luksan47/mars/issues/125

### LINKED ISSUE #689: Printing from `DocumentController` is broken

horcsinbalint:
The changes allow us to have multiple printers and select the destination on the web interface.
It has been implemented in a way that seperate PrinterOptions can be created for setting different valid settings for printing (like single / double sided printing).
It also gives enormous flexibility for system administrators as it no longer needs additional code in mars to allow new methods of printing (like flip on short edge option is simply just a new configuration with a different flags given to lp). Some combinations (borderless, colored, duplex) are incompatible, developing these rules would be extremely difficult and depends on the currently available machines so I think listing all available options is the best way of making sure that all available options are supported.
Sometimes we already needed to disable double-sided printing temporarily. This PR also offers a clean solution for it.
Additionally, it fixes #376, #689

Calling `lpstat` every time the printing page is loaded is extremely time consuming and not even that useful. Moving it to its own seperate job that runs every 30 seconds.
Therefore fixes #375


resources/lang/en/document.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+return [
+    'departure-statement' => 'Departure statement',
+    'download' => 'Download',
+    'fill' => 'Fill',
+    'import' => 'Import license',
+    'printing_first' => 'You can print documents here, and the cost will be deducted from your printing account (printing configuration: :printing_conf_name_eng, your balance: HUF :current_balance). If you want to use different settings (e.g., use free printing credits or print multiple copies), download the document and upload it on the ',
+    'printing_second' => 'printing page',
+    'printing_third' => '.',
```
BertalanD: nit: this doesn't look ideal for maintainability, etc. How about just including the `a` tag in the string, and then  passing the URL as a variable to `__`?

resources/views/dormitory/print/print.blade.php
```diff
@@ -28,48 +27,67 @@
...
+            </form>
+        @endforeach
         <form class="form-horizontal" role="form" method="POST" action="{{ route('print.print-job.store') }}" enctype="multipart/form-data">
             @csrf
+            @foreach($printers as $index => $printer)
+                <div class="row">
+                    <label style="padding-left: 25px; margin: 5px">
+                        <input type="radio" name="printer" value="{{ $printer->id }}"
+                            {{ $index === 0 ? 'checked' : '' }} {-- Default selection --}
+                        >
```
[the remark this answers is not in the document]
BertalanD: this looks relevant

resources/lang/en/document.php
horcsinbalint: I also wondered how it could be done. But I didn't want to add the `<a>` inside the language file as `:printing_conf_name_eng` is loaded from the database which could potentially be a step during an XSS attack, that's why I put it inside `{{}}` for sanitization (and that would also destroy the `<a>` tag). We could probably use `@Markdown` in the blade to allow formatted text. I will make a commit for it in a bit.

BertalanD:
Last nit question, otherwise LGTM: why was the change to the button component necessary?

Before:
image

After:
image

In other cases where there are multiple buttons on one card, they aren't stretched to the full width:
image

horcsinbalint:
> Last nit question, otherwise LGTM: why was the change to the button component necessary?

On some places the buttons were misaligned without this change.

Issue #701, opened 2025-02-03
Wrong CTSzK reference in end-of-semester form


horcsinbalint:

horcsinbalint:
Fixed via #736

Issue #698, opened 2025-02-01
General assembly attendance is misleading in semester evaluation form


horcsinbalint:
![Image](https://github.com/user-attachments/assets/9ab14cf2-af82-4ba1-8dac-58e01d3fa427)
It is shown that the user has attended if they were excused but were not present. It is a bit misleading

horcsinbalint:
Fixed via #736

PR #736, opened 2025-06-06
Fix semester evaluation form

horcsinbalint:
There have been some issues with CTSzK references (#701) and misleading text regarding general assembly attendance (#698). This commit should fix this.

Another issue was that the delete button did not work for anonymous questions. It should be fixed now.


kamillacziraki:
All the changes in the texts look good to me, although it seems that in one place, the CTSzK reference stayed the same wrong one (7. § instead of 8. §). It is where the consequences for not having the 4.25 GPA are described; the text was changed, but the reference stayed the same (line 12 of the resources/views/secretariat/evaluation-form/avg.blade.php).

horcsinbalint:
> All the changes in the texts look good to me, although it seems that in one place, the CTSzK reference stayed the same wrong one (7. § instead of 8. §). It is where the consequences for not having the 4.25 GPA are described; the text was changed, but the reference stayed the same (line 12 of the resources/views/secretariat/evaluation-form/avg.blade.php).

Thanks for catching it.

PR #744, opened 2025-07-09
Fix Alfonso tests for the spring semester

BertalanD:
The function that checks whether a user has been admitted as a bachelor student assumes that admittance semesters are autumn semesters. This is a fine thing to do in my opinion, but the tests didn't account for that and therefore were broken when we ran them in the spring.

This resolves the two unit test failures on CI.

Fixes #629


horcsinbalint (approved):
I was wondering why the CI was failing for the last couple of months.
Good catch, thanks for checking it.

Issue #742, opened 2025-07-07
Prepare for this year's applications


horcsinbalint:
- [x] Rename tabs
- [x] Better description texts:
   - [x] Allow MD (?)
- [x] Allow saving when only parts of the data is filled
  - [x] only check if these are present during finalization
- [x] Blur / disable pictures by default
- [x] Allow opening / finalizing applications by sysadmins, allow edits by sysadmins
  - [x] Sysadmins should be able to view profile pages of unverified users
- [x] Different document types (like érettségi bizonyítvány, besorolási határozat, etc.)
- [ ] Suggest logging in on `/register` if an account exists
- [x] Add opt out for publishing names on the website (+ codeword field)

PR #752, opened 2025-07-23
Distinguish different document types for applications

BertalanD:
- Introduce a type enum in the files table
- Add a dedicated section in the UI for each required document type
- Add validation rules that check if each required document type was
  provided

The description texts in the UI can be changed via the configurable
texts framework without re-deploying the code.

Work towards #742

Sorry for the painful Hunglish, but I'm not going to try to translate the official document type names to English.

### DIFF
--- a/app/Http/Controllers/Secretariat/UserController.php
+++ b/app/Http/Controllers/Secretariat/UserController.php
@@ -69,7 +69,7 @@ class UserController extends Controller
             $old_profile->update(['path' => $path]);
             Storage::delete($old_profile->path);
         } else {
-            $user->profilePicture()->create(['path' => $path, 'name' => 'profile_picture']);
+            $user->profilePicture()->create(['path' => $path, 'type' => 'profile_picture']);
         }
         return redirect()->back()->with('message', __('general.successful_modification'));
     }
--- a/app/Models/Application.php
+++ b/app/Models/Application.php
@@ -3,6 +3,7 @@
 namespace App\Models;
 
 use App\Utils\DataCompresser;
+use App\Enums\FileType;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Casts\Attribute;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
@@ -171,6 +172,15 @@ class Application extends Model
         return $this->hasMany('App\Models\File');
     }
 
+    /**
+     * Get files of a specific type.
+     * @param FileType $type
+     * @return HasMany
+     */
+    public function filesOfType(FileType $type): HasMany
+    {
+        return $this->files()->where('type', $type);
+    }
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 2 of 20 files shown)


app/Http/Controllers/Secretariat/UserController.php
```diff
@@ -69,7 +69,7 @@ public function storeProfilePicture(Request $request, User $user): RedirectRespo
             $old_profile->update(['path' => $path]);
             Storage::delete($old_profile->path);
         } else {
-            $user->profilePicture()->create(['path' => $path, 'name' => 'profile_picture']);
+            $user->profilePicture()->create(['path' => $path, 'type' => 'profile_picture']);
```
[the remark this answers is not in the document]
horcsinbalint: AI's suggestion is not wrong, but does not really matter

app/Models/Application.php
```diff
@@ -413,13 +423,37 @@ public function missingData(): array
...
+     * @param FileType $type
+     * @return bool
+     */
+    public function needsFile(FileType $type): bool
+    {
+        switch ($type) {
+            case FileType::RESUME:
+                return true;
+            case FileType::BESOROLASI_HATAROZAT:
+                return $this->user->educationalInformation->studyLines()->where('start', '=', Semester::next()->id)->exists();
```
horcsinbalint: I think we should probably use `app(\App\Http\Controllers\Auth\ApplicationController::class)->semester` for getting the current semester as the application is a periodic event

app/Utils/ApplicationHandler.php
```diff
@@ -70,9 +71,10 @@ public function storeFile(Request $request, $user): void
         $request->validate([
             'file' => 'required|file|mimes:pdf,jpg,jpeg,png|max:' . config('custom.general_file_size_limit'),
             'name' => 'required|string|max:255',
+            'type' => ['required', Rule::enum(\App\Enums\FileType::class)],
         ]);
         $path = $request->file('file')->store('uploads');
-        $user->application->files()->create(['path' => $path, 'name' => $request->input('name')]);
+        $user->application->files()->create(['path' => $path, 'type' => $request->input('type'), 'description' => $request->input('name')]);
```
horcsinbalint, line 80: Just a question: is it safe to allow users to upload any type of document (like a `receipt`)?

resources/views/auth/application/application.blade.php
```diff
@@ -306,7 +306,7 @@ class="hideable-picture"
                                         <div class="row" style="margin-bottom: 0; padding: 10px">
                                             <div class="col" style="margin-top: 5px">
                                                 <a href="{{ url($file->path) }}"
-                                                   target="_blank">{{ $file->name }}</a>
+                                                   target="_blank">{{ $file->description }}</a>
                                             </div>
                                         </div>
                                     @empty
```
horcsinbalint, line 313: I think line #317 is no longer necessary.

horcsinbalint (commented):
Thank you, the changes seem to be quite good, I added some nitpicking comments

app/Utils/ApplicationHandler.php
BertalanD: Hm, I don't think that has any security implications. They will still have their `application` foreign key set, so will be visible in the admission UI and will be deleted after the admissions process finishes.

resources/sass/materialize.scss
```diff
@@ -390,3 +390,11 @@ body.dark a.btn {
     width: 100%;
     text-align: center;
 }
+
+.form-disabled {
+    opacity: 0.5;
+
+    & .collapsilble-body {
```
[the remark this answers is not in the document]
BertalanD: I'm just removing that. We should still be able to interact with the dropdown and the file upload/delete buttons.

horcsinbalint (approved):

PR #755, opened 2025-07-23
Do not allow workshop administrators to add application committee role

horcsinbalint:


BertalanD (approved):
This is horrible, but it works.

PR #761, opened 2025-07-28
Pre-fill averages for two semesters

BertalanD:
By creating two input boxes when the user selects the "I have previous finished semesters" check box, we can avoid people creating a single entry by taking the average across all semesters.

Obligatory "this is a hack".


horcsinbalint (changes requested):
It checks the checkbox by default, it would be better if there was a default value option that is set if it was previously unfilled but the checkbox is checked.

BertalanD:
can't approve as I'm the creator of this PR, but your fix LGTM

PR #765, opened 2025-07-30
De-duplicate available workshops on /applicants

BertalanD:
Without this change, if a user is both an administrator for a workshop and has an application committee role for it, the workshop's name would show up twice in the workshop chooser dropdown.

Reported by Tóth Fanni.


BertalanD:
Merging with verbal approval from @horcsinbalint

PR #767, opened 2025-08-04
Fix phone number regexes

BertalanD:
HTML treats input validation patterns as v-mode RegExes, which requires special characters to be escaped in more cases. The previous patterns were syntactically invalid, and thus ignored by browsers:

```
Pattern attribute value [+][0-9]{1,4}[-\s()0-9]* is not a valid regular
expression:

Uncaught SyntaxError: Invalid regular expression:
/[+][0-9]{1,4}[-\s()0-9]*/v: Invalid character in character class
```

### DIFF
--- a/resources/views/auth/register.blade.php
+++ b/resources/views/auth/register.blade.php
@@ -20,7 +20,7 @@
                     <div class="markdown_with_red">
                         @markdown(\App\Models\ConfigurableText::getText("APPLICANT_REGISTRATION"))
                     </div>
-                    
+
                     @else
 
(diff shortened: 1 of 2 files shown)


horcsinbalint:
I'd only fix this after this year's application period as it would be quite confusing for applicants if they edit some other personal information and the system just starts randomly asking them to edit their phone number. There is only one phone number that is actually unusable for non-verified users. I would probably go through manually on the users' data - many start with 06 or 07 - and fix it for them after this PR is merged (but I would only do it after the application period).

PR #806, opened 2026-07-01
Application 2026

horcsinbalint:
Prepare Mars's functional changes for the 2026 application process.

Changes:
- notes for the committees were just a burden in this form, there were external tools used mostly for communication within the committees, removed the feature
- a motivational letter is now requested for each applied workshop


horcsinbalint:
Merging, as it does not effect the current production and post-commit review can be done. We can revert these changes or apply patches afterwards.

